PR #507, opened 2020-08-20
Catch extra exceptions when server connection closes

cliffburdick:
Copied from here:
https://github.com/zalando-incubator/kopf/pull/386

Catches more exceptions that can happen when the server shuts down.

## Description

Currently the only exception that is caught when a connection to the server fails is aiohttp.ClientResponseError. However, we've observed two others occurring:

aiohttp.client_exceptions.ServerDisconnectedError
aiohttp.client_exceptions.ClientOSError

This PR catches them, and handles them the same way as the previous error.

## Issues/PRs

> Issues: 
#204 

> Related:

## Type of changes

## Checklist

- [X] The code addresses only the mentioned problem, and this problem only
- [X] I think the code is well written

### DIFF
--- a/kopf/clients/events.py
+++ b/kopf/clients/events.py
@@ -94,3 +94,4 @@ async def post_event(
     except aiohttp.ClientOSError as e:
         logger.warning(f"Failed to post an event. Ignoring and continuing. "
                        f"Event: type={type!r}, reason={reason!r}, message={message!r}.")
+                       
--- a/kopf/clients/events.py
+++ b/kopf/clients/events.py
@@ -81,9 +81,16 @@ async def post_event(
         )
         response.raise_for_status()
 
-    except (aiohttp.ClientResponseError, aiohttp.client_exceptions.ServerDisconnectedError, aiohttp.client_exceptions.ClientOSError) as e:
-        # Events are helpful but auxiliary, they should not fail the handling cycle.
-        # Yet we want to notice that something went wrong (in logs).
+    # Events are helpful but auxiliary, they should not fail the handling cycle.
+    # Yet we want to notice that something went wrong (in logs).
+    except aiohttp.ClientResponseError as e:
         logger.warning(f"Failed to post an event. Ignoring and continuing. "
                        f"Status: {e.status}. Message: {e.message}. "
+                       f"Event: type={type!r}, reason={reason!r}, message={message!r}.")
+    except aiohttp.ServerDisconnectedError as e:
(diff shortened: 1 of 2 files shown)


kopf/clients/events.py
```diff
@@ -81,7 +81,7 @@ async def post_event(
         )
         response.raise_for_status()
 
-    except aiohttp.ClientResponseError as e:
+    except (aiohttp.ClientResponseError, aiohttp.client_exceptions.ServerDisconnectedError, aiohttp.client_exceptions.ClientOSError) as e:
```
nolar: If the connection is not reliable, the SSL handshake can also fail.

Generally, I suggest ignoring absolutely any errors in k8s-event posting, up to the very base `aiohttp.ClientError`. 

**Why?** ==> By design, k8s-event posting in a background task (i.e. not interactive) is a supplementary feature, so should not break the operator in any case.  In addition, the errors cannot be delivered back to the stack frame from which the event was posted — because of its background nature (the original stack trace only puts a message to the queue, and does not post anything). So, it is a fire-and-forget principle.

**A side-note:** Later, retries can be added for k8s-events: i.e. up to 3 times (or a configurable number of times) before really giving up and logging the failure. But this will be an enhancement, not a bug fix. I've extracted this into #513.

For now, just ignoring (but logging) any client-side errors is fine.

**The PR note No 1 (would-be-bug):** When implementing, please take into account that `ClientResponseError` here has `.status` & `.message`, which are used in the log message on the next line. `ServerDisconnectedError` has only `.message`, but not `.status`. `ClientOSError` has neither of them.

**The PR note No 2 (tech-debt):** Also note that aiohttp's errors are exported as `aiohttp.ServerDisconnectedError` (as per [aiohttp's documentation](https://docs.aiohttp.org/en/stable/client_reference.html#client-exceptions)), not as `aiohttp.client_exceptions…`.

**The PR note No 3 (an improvement):** It might be a good idea to have several `except` clauses here, starting from `ClientResponseError`, and then the base `ClientError` — with different log messages for each, including different levels of details available from the error: i.e. for ClientResponseErrors, I would like to see the status for sure; for disconnects, not so much.

nolar (changes requested):

kopf/clients/events.py
cliffburdick: Thanks @nolar! I didn't notice the different types, so I will change that and resubmit the PR.

nolar (approved):

PR #508, opened 2020-08-20
Fix #384: Prune sub-handler state when done

### LINKED ISSUE #384: Subhandler states are not cleared at the end of handling cycle
> <a href="https://github.com/MarkusH"></a>        An issue by [MarkusH](https://github.com/MarkusH) at _2020-07-11 16:29:25+00:00_
> Original URL: https://github.com/zalando-incubator/kopf/issues/384
> &nbsp;

## Long story short

When using `kopf.register()` or `kopf.execute()`, the state of the "root" handler is tracked in `kopf.reactor.processing.process_resource_changing_cause()` which calls [`state.store()`](https://github.com/nolar/kopf/blob/349a976f65599832ac91db6bc91d14365bf497ba/kopf/reactor/processing.py#L381). It also calls [`state.purge()`](https://github.com/nolar/kopf/blob/349a976f65599832ac91db6bc91d14365bf497ba/kopf/reactor/processing.py#L386) a few lines below, when `state.done`. For sub-handlers, there is a [`state.store()`](https://github.com/nolar/kopf/blob/349a976f65599832ac91db6bc91d14365bf497ba/kopf/reactor/handling.py#L170) call in `kopf.reactor.handling.execute()`. However, the state for sub-handlers is never purged.

## Description

The failing purging of sub-handler states leads to sub-handlers not being executed. I'm not sure if that has implications or may even be the cause for #279.

```python
import functools

import kopf


async def subhandler1(**kwargs):
    pass


async def subhandler2(**kwargs):
    pass


@kopf.on.update("example.com", "v1", "as")
async def update(**kwargs):
    await kopf.execute(
        fns={"s1": functools.partial(subhandler1), "s2": functools.partial(subhandler2)}
    )
```

```yaml
apiVersion: example.com/v1
kind: A
metadata:
  annotations:
    kopf.zalando.org/last-handled-configuration: '{"spec": {"a": 1}}'
  creationTimestamp: "2020-07-11T16:19:28Z"
  generation: 1
  name: test
  namespace: default
  resourceVersion: "4657318"
  selfLink: /apis/example.com/v1/namespaces/default/as/test
  uid: a79a6ca3-687c-4467-abc1-a38f0019a5ca
spec:
  a: 1
```

After triggering a change:

```yaml
apiVersion: example.com/v1
kind: A
metadata:
  annotations:
    kopf.zalando.org/last-handled-configuration: '{"spec": {"a": 2}}'
    kopf.zalando.org/update.s1: '{"started": "2020-07-11T16:20:02.463066", "stopped":
      "2020-07-11T16:20:02.464577", "retries": 1, "success": true, "failure": false}'
    kopf.zalando.org/update.s2: '{"started": "2020-07-11T16:20:02.463137", "stopped":
      "2020-07-11T16:20:02.737845", "retries": 1, "success": true, "failure": false}'
  creationTimestamp: "2020-07-11T16:19:28Z"
  generation: 4
  name: test
  namespace: default
  resourceVersion: "4657369"
  selfLink: /apis/example.com/v1/namespaces/default/as/test
  uid: a79a6ca3-687c-4467-abc1-a38f0019a5ca
spec:
  a: 2
status:
  kopf:
    progress:
      update/s1:
        failure: false
        retries: 1
        started: 2020-07-11T16:20:02.463066
        stopped: 2020-07-11T16:20:02.464577
        success: true
      update/s2:
        failure: false
        retries: 1
        started: 2020-07-11T16:20:02.463137
        stopped: 2020-07-11T16:20:02.737845
        success: true
```

```
[2020-07-11 18:20:02,460] kopf.objects         [DEBUG   ] [default/test] Update event: {'apiVersion': 'example.com/v1', 'kind': 'A', 'metadata': {'annotations': {'kopf.zalando.org/last-handled-configuration': '{"spec": {"a": 1}}'}, 'creationTimestamp': '2020-07-11T16:19:28Z', 'generation': 2, 'name': 'test', 'namespace': 'default', 'resourceVersion': '4657365', 'selfLink': '/apis/example.com/v1/namespaces/default/as/test', 'uid': 'a79a6ca3-687c-4467-abc1-a38f0019a5ca'}, 'spec': {'a': 2}}
[2020-07-11 18:20:02,460] kopf.objects         [DEBUG   ] [default/test] Update diff: (('change', ('spec', 'a'), 1, 2),)
[2020-07-11 18:20:02,462] kopf.objects         [DEBUG   ] [default/test] Handler 'update' is invoked.
[2020-07-11 18:20:02,463] kopf.objects         [DEBUG   ] [default/test] Handler 'update/s1' is invoked.
[2020-07-11 18:20:02,464] kopf.objects         [INFO    ] [default/test] Handler 'update/s1' succeeded.
[2020-07-11 18:20:02,465] kopf.objects         [DEBUG   ] [default/test] Handler 'update' has unfinished sub-handlers. Will retry soon.
[2020-07-11 18:20:02,465] kopf.objects         [DEBUG   ] [default/test] Patching with: {'metadata': {'annotations': {'kopf.zalando.org/update.s1': '{"started": "2020-07-11T16:20:02.463066", "stopped": "2020-07-11T16:20:02.464577", "retries": 1, "success": true, "failure": false}', 'kopf.zalando.org/update.s2': '{"started": "2020-07-11T16:20:02.463137", "retries": 0, "success": false, "failure": false}', 'kopf.zalando.org/update': '{"started": "2020-07-11T16:20:02.461880", "delayed": "2020-07-11T16:20:02.465470", "retries": 1, "success": false, "failure": false, "message": "None"}'}}, 'status': {'kopf': {'progress': {'update/s1': {'started': '2020-07-11T16:20:02.463066', 'stopped': '2020-07-11T16:20:02.464577', 'delayed': None, 'retries': 1, 'success': True, 'failure': False, 'message': None}, 'update/s2': {'started': '2020-07-11T16:20:02.463137', 'stopped': None, 'delayed': None, 'retries': 0, 'success': False, 'failure': False, 'message': None}, 'update': {'started': '2020-07-11T16:20:02.461880', 'stopped': None, 'delayed': '2020-07-11T16:20:02.465470', 'retries': 1, 'success': False, 'failure': False, 'message': 'None'}}}}}
[2020-07-11 18:20:02,733] kopf.objects         [DEBUG   ] [default/test] Update event: {'apiVersion': 'example.com/v1', 'kind': 'A', 'metadata': {'annotations': {'kopf.zalando.org/last-handled-configuration': '{"spec": {"a": 1}}', 'kopf.zalando.org/update': '{"started": "2020-07-11T16:20:02.461880", "delayed": "2020-07-11T16:20:02.465470", "retries": 1, "success": false, "failure": false, "message": "None"}', 'kopf.zalando.org/update.s1': '{"started": "2020-07-11T16:20:02.463066", "stopped": "2020-07-11T16:20:02.464577", "retries": 1, "success": true, "failure": false}', 'kopf.zalando.org/update.s2': '{"started": "2020-07-11T16:20:02.463137", "retries": 0, "success": false, "failure": false}'}, 'creationTimestamp': '2020-07-11T16:19:28Z', 'generation': 3, 'name': 'test', 'namespace': 'default', 'resourceVersion': '4657366', 'selfLink': '/apis/example.com/v1/namespaces/default/as/test', 'uid': 'a79a6ca3-687c-4467-abc1-a38f0019a5ca'}, 'spec': {'a': 2}}, 'status': {'kopf': {'progress': {'update': {'delayed': '2020-07-11T16:20:02.465470', 'failure': False, 'message': 'None', 'retries': 1, 'started': '2020-07-11T16:20:02.461880', 'success': False}, 'update/s1': {'failure': False, 'retries': 1, 'started': '2020-07-11T16:20:02.463066', 'stopped': '2020-07-11T16:20:02.464577', 'success': True}, 'update/s2': {'failure': False, 'retries': 0, 'started': '2020-07-11T16:20:02.463137', 'success': False}}}}}
[2020-07-11 18:20:02,734] kopf.objects         [DEBUG   ] [default/test] Update diff: (('change', ('spec', 'a'), 1, 2),)
[2020-07-11 18:20:02,735] kopf.objects         [DEBUG   ] [default/test] Handler 'update' is invoked.
[2020-07-11 18:20:02,736] kopf.objects         [DEBUG   ] [default/test] Handler 'update/s2' is invoked.
[2020-07-11 18:20:02,737] kopf.objects         [INFO    ] [default/test] Handler 'update/s2' succeeded.
[2020-07-11 18:20:02,738] kopf.objects         [INFO    ] [default/test] Handler 'update' succeeded.
[2020-07-11 18:20:02,739] kopf.objects         [INFO    ] [default/test] All handlers succeeded for update.
[2020-07-11 18:20:02,740] kopf.objects         [DEBUG   ] [default/test] Patching with: {'metadata': {'annotations': {'kopf.zalando.org/update.s2': '{"started": "2020-07-11T16:20:02.463137", "stopped": "2020-07-11T16:20:02.737845", "retries": 1, "success": true, "failure": false}', 'kopf.zalando.org/update': None}}, 'status': {'kopf': {'progress': {'update/s2': {'started': '2020-07-11T16:20:02.463137', 'stopped': '2020-07-11T16:20:02.737845', 'delayed': None, 'retries': 1, 'success': True, 'failure': False, 'message': None}, 'update': None}}}}
[2020-07-11 18:20:02,883] kopf.objects         [DEBUG   ] [default/test] Something has changed, but we are not interested (the essence is the same).
[2020-07-11 18:20:02,883] kopf.objects         [DEBUG   ] [default/test] Handling cycle is finished, waiting for new changes since now.
```

Making yet another update to the custom resource:

```yaml
apiVersion: example.com/v1
kind: A
metadata:
  annotations:
    kopf.zalando.org/last-handled-configuration: '{"spec": {"a": 3}}'
    kopf.zalando.org/update.s1: '{"started": "2020-07-11T16:20:02.463066", "stopped":
      "2020-07-11T16:20:02.464577", "retries": 1, "success": true, "failure": false}'
    kopf.zalando.org/update.s2: '{"started": "2020-07-11T16:20:02.463137", "stopped":
      "2020-07-11T16:20:02.737845", "retries": 1, "success": true, "failure": false}'
  creationTimestamp: "2020-07-11T16:19:28Z"
  generation: 5
  name: test
  namespace: default
  resourceVersion: "4748091"
  selfLink: /apis/example.com/v1/namespaces/default/as/test
  uid: a79a6ca3-687c-4467-abc1-a38f0019a5ca
spec:
  a: 3
status:
  kopf:
    progress:
      update/s1:
        failure: false
        retries: 1
        started: 2020-07-11T16:20:02.463066
        stopped: 2020-07-11T16:20:02.464577
        success: true
      update/s2:
        failure: false
        retries: 1
        started: 2020-07-11T16:20:02.463137
        stopped: 2020-07-11T16:20:02.737845
        success: true
```

```
[2020-07-12 11:46:28,666] kopf.objects         [DEBUG   ] [default/test] Update event: {'apiVersion': 'example.com/v1', 'kind': 'A', 'metadata': {'annotations': {'kopf.zalando.org/last-handled-configuration': '{"spec": {"a": 2}}', 'kopf.zalando.org/update.s1': '{"started": "2020-07-11T16:20:02.463066", "stopped": "2020-07-11T16:20:02.464577", "retries": 1, "success": true, "failure": false}', 'kopf.zalando.org/update.s2': '{"started": "2020-07-11T16:20:02.463137", "stopped": "2020-07-11T16:20:02.737845", "retries": 1, "success": true, "failure": false}'}, 'creationTimestamp': '2020-07-11T16:19:28Z', 'generation': 5, 'name': 'test', 'namespace': 'default', 'resourceVersion': '4748090', 'selfLink': '/apis/example.com/v1/namespaces/default/as/test', 'uid': 'a79a6ca3-687c-4467-abc1-a38f0019a5ca'}, 'spec': {'a': 3}}, 'status': {'kopf': {'progress': {'update/s1': {'failure': False, 'retries': 1, 'started': '2020-07-11T16:20:02.463066', 'stopped': '2020-07-11T16:20:02.464577', 'success': True}, 'update/s2': {'failure': False, 'retries': 1, 'started': '2020-07-11T16:20:02.463137', 'stopped': '2020-07-11T16:20:02.737845', 'success': True}}}}}
[2020-07-12 11:46:28,666] kopf.objects         [DEBUG   ] [default/test] Update diff: (('change', ('spec', 'a'), 2, 3),)
[2020-07-12 11:46:28,668] kopf.objects         [DEBUG   ] [default/test] Handler 'update' is invoked.
[2020-07-12 11:46:28,669] kopf.objects         [INFO    ] [default/test] Handler 'update' succeeded.
[2020-07-12 11:46:28,670] kopf.objects         [INFO    ] [default/test] All handlers succeeded for update.
[2020-07-12 11:46:28,671] kopf.objects         [DEBUG   ] [default/test] Patching with: {'metadata': {'annotations': {'kopf.zalando.org/last-handled-configuration': '{"spec": {"cluster": {"imageRegistry": "crate", "license": {"secretKeyRef": {"key": "license-key", "name": "cratedb-license"}}, "name": "my-crate-cluster", "version": "4.1.5"}, "nodes": {"data": [{"name": "t1", "replicas": 8, "resources": {"cpus": 0.25, "disk": {"count": 1, "size": "8GiB", "storageClass": "default"}, "heapRatio": 0.5, "memory": "512Mi"}}]}}}'}}}
[2020-07-12 11:46:28,921] kopf.objects         [DEBUG   ] [default/test] Something has changed, but we are not interested (the essence is the same).
[2020-07-12 11:46:28,921] kopf.objects         [DEBUG   ] [default/test] Handling cycle is finished, waiting for new changes since now.
```

## Environment

* Kopf version: 0.27

---

> <a href="https://github.com/nemethf"></a>            Commented by [nemethf](https://github.com/nemethf) at _2020-07-27 12:12:15+00:00_
> &nbsp;

I think I ran into this as well.  My workaround was to append kwarg["started"] to the id of the kopf.on.this decorator:
```python
        @kopf.on.this(id=f's1-kwargs["started"]')
        async def subhandler1(**kwargs):
            pass
```
With this workaround, the sub-hander runs on each update, but kopf now litters the annotations of the custom resource. I hope PR #385 fixes this issue.  Thanks, [MarkusH](https://github.com/MarkusH), for that.

---

> <a href="https://github.com/MarkusH"></a>            Commented by [MarkusH](https://github.com/MarkusH) at _2020-08-11 17:55:03+00:00_
> &nbsp;

Hi [nolar](https://github.com/nolar), I can only imagine how busy you must be. Would you be able to look into the PR (#385) and let me know if the approach taken there makes sense? I'm happy to invest more time to get this problem fixed. Thank you. :slightly_smiling_face:

MarkusH:
Without this change, the completed state remains in storage, causing the
handlers to not be executed again in a new handling cycle.

## Issues/PRs

Issues: #384 

Related: #279 

## Type of changes

## Checklist

### DIFF
--- a/kopf/reactor/handling.py
+++ b/kopf/reactor/handling.py
@@ -170,8 +170,10 @@ async def execute(
     state.store(body=cause.body, patch=cause.patch, storage=storage)
     states.deliver_results(outcomes=outcomes, patch=cause.patch)
 
-    # Escalate `HandlerChildrenRetry` if the execute should be continued on the next iteration.
-    if not state.done:
+    if state.done:
+        state.purge(body=cause.body, patch=cause.patch, storage=storage)
+    else:
+        # Escalate `HandlerChildrenRetry` if the execute should be continued on the next iteration.
         raise HandlerChildrenRetry(delay=state.delay)
 
 


nolar:
@MarkusH Thanks for this issue reported and this PR. Indeed, the sub-handlers are not purged now.  I'll try to write down my thoughts on this:

**For historic context:** 

Previously, they were purged because the whole `.status.kopf` section was purged, regardless of the content. But since the introduction of state/progress storages, and especially the annotations storage (which was needed because of the recently introduced K8s 1.16+ "structural schemas", which made status nearly unusable, #331 & co) they now care only about their own keys, and skip the "unknown" keys, as they can belong to other operators.

**Unsolved cases:**

Indeed, the keys of sub-handlers are now considered as "unknown" keys, which is not intended. However, the fix in this PR solves it partially, but not fully. There are other cases when it will introduce another kind of misbehavior:

It will work if the sub-handlers are the last thing done in the main handler, and the main handler can never fail after the sub-handlers have succeeded — which is surely the case with `kopf.register()`, and the cases with `kopf.execute()` in the end of the main handler.

If a handler calls the sub-handlers via `kopf.execute()`, and then does something else afterward in an assumption that the sub-handlers are done, and that "something" fails, the main handler will fail and will be retried. To the moment when it is retried, the information about the sub-handlers' success will be purged, so the sub-handlers will be re-executed anew. This is also not intended.

**Why is this important:**

The idea of all handlers (sub- & regular) is kind of "execute once [per cause]" approach. For example, we used it for creating children resources, such as Job, PVCs, etc, and be sure that if the object creation succeeds, it is never tried again.

The sub-handlers are also intended to be dynamically decided: on every call to the main handler, there can be a different set of sub-handlers based on conditions and external circumstances. If the set of sub-handlers is fixed, they should be better be regular handlers — it is simpler and less complicated.

Kopf's internal behavior is to purge the state when we are sure that everything is done (succeeded or failed permanently) and the eventual consistency (of handlers-vs-object) is reached. Both the purge of the state and the preservation of the last-handled-state is done in one `PATCH` operation after the last handler succeeds — and this operation is atomic (or close to it). If it is done in different places, this breaks the "execute once" promise.

**This PR's solution** can be extended: first, purge the sub-handlers' states in the main handler, then also mark the main handler as to never try the sub-handlers again. 

This would be not a valid assumption from the framework side, and also breaks the dynamic nature of the sub-handlers.

**Another way** would be to purge the state so that all involved sub-handlers are also purged. The tricky part is that, at the moment of the purge, we don't know which sub-handlers exist, as we most likely didn't execute the main handler on this iteration. But we know the top-level handlers' ids. So, here, we can only _assume_ that all the persistence keys that _start with_ the main handlers' ids are also to be purged.

This presumably does not break the support of overlapping operators for the same resource, but with different handler ids: if they happen to have different sub-handlers' ids, but the same root handlers' ids, they are already misdesigned and would cause trouble anyway.

So, the best solution for this problem seems to be somewhere in the scope of `state`<->`storage` interaction and their `purge()` methods: the `key` must not be taken directly, but as a prefix of all keys to purge (potentially with knowing the structure of the keys (storage's internal logic), not just as strings). Speaking in SQL terms for the sake of mnemonics, it should not be `DELETE … WHERE key = {key}`, but `DELETE … WHERE key = "{key}" OR key LIKE "{key}.%"`.

**And looking forward:** I do not see a reason to operate on individual keys: the purge is done for the whole state, multiple involved keys. The per-key iteration was probably done this way in some wrong assumptions. Imagine an SQL-like storage: we can delete all the needed keys in one `DELETE` query, not in a query-per-key-prefix multiplied by the number of keys.

I'll try to look into it now…

nolar:
PS: … But we cannot filter the existing keys by prefix, as in the annotations storage, the keys can be hashed before the first dot — to fit into K8s's 63-chat limit for annotation/label names. So, the actual names of the keys are not known in the storage. 

Which leads us back to the solution with purging in `kopf.execute()`, and the need to handle the retries of the main handler or a case with 2+ `kopf.execute()` calls.

This issue is not easy, indeed…

MarkusH:
Hi @nolar I appreciate your in-depth input in this topic. I've yet to give your PR a try, but does it work with sub-sub-handlers? I.e a handler that calls `execute()` and some of the sub-handlers calling `execute()` as well?

nolar:
@MarkusH Yes. That is even shown in the PR description with "update/s2/z1" and alike "z"-handlers (the second example there). To my shame, I barely can imagine a realistic use-case for that (based on my experience), but at least the hypothetical behavior is consistent for all handlers — root & sub — after the fix. 

I am yet going to separately test it with the case of lengthy handler ids and/or operator's identity prefix — so that the annotation names are suffix-hashed to fit into a 63-character limit, but the state remains purgeable as it should.

PR #518, opened 2020-08-23
Select the field-handlers based on actual diff of that field present

nolar:
## What do these changes do?

Stop invoking the field handlers for _non-existent_ fields on the creation of the objects and on big changes with structures that are supposed to contain those fields, but they don't.

## Description

**Historic context:** Filtering only by diff's fields without actual diff values checked was used since the initial import and release of the framework, and was only modified in #340 by extending the filtering logic to react to sub-fields of a bigger structure changed (for example, for a field "spec.struct.field" when "spec.struct" is added at once).

**Problem:** Yet such an approach is prone to improper detection of actually changed fields: when a bigger struct is added while we expect a non-existent sub-field of it, it is believed to exist there, as the sub-field matches the bigger struct; in particular, when an object is created, it is a root-level diff from `None` to the actual content, and all the non-existent fields trigger their field handlers.

A sample way to reproduce with the existing `examples/obj.yaml` (xxx-handler should not be called at all):

```python
import kopf

@kopf.on.field('zalando.org', 'v1', 'kopfexamples', field='spec.xxx')
def updateField_changed(old, new, **_):
    print(f"XXX! {old!r} -> {new!r}")

@kopf.on.field('zalando.org', 'v1', 'kopfexamples', field='spec.field')
def updateField_changed(old, new, **_):
    print(f"FIELD! {old!r} -> {new!r}")
```

**Solution:** By checking if the actual diff of a specific field is not empty, we can properly detect if that specific field was added, removed, or changed, be that as a part of a bigger (containing) structure change, or as a change in any of its (contained) sub-fields.

**Trade-offs:** This comes at a cost of performance loss: the diff reduction is less performant than a flat iteration over the diff fields: from 20% to 150% (2.5x) slower. However, the scale of the slowdown is measured in 5-10s per 1'000'000 checks on MacBook Air 2013, so we can consider it negligible for real operators.

**Important!** To be released as 0.27.1 hotfix, and later merged into the master branch (#???).

## Issues/PRs

> Issues: #375 

> Related: #340 

## Type of changes

## Checklist

- [x] Documentation reflects the changes

### DIFF
--- a/kopf/reactor/registries.py
+++ b/kopf/reactor/registries.py
@@ -16,7 +16,7 @@ import collections
 import functools
 import warnings
 from types import FunctionType, MethodType
-from typing import (Any, MutableMapping, Optional, Sequence, Collection, Iterable, Iterator,
+from typing import (Any, MutableMapping, Optional, Sequence, Iterable, Iterator,
                     List, Set, FrozenSet, Mapping, Callable, cast, Generic, TypeVar, Union,
                     Container)
 
@@ -24,6 +24,7 @@ from kopf.reactor import causation
 from kopf.reactor import invocation
 from kopf.structs import callbacks
 from kopf.structs import dicts
+from kopf.structs import diffs
 from kopf.structs import filters
 from kopf.structs import handlers
 from kopf.structs import resources as resources_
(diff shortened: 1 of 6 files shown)


nolar:
Closed in favor of #523 — with a proper base branch (master, not release/0.27).

PR #551, opened 2020-09-15
Introduce pre-commit hooks and Github Actions

turbaszek:
## Description

Introduce pre-commits to automate running static check:
https://pre-commit.com

## Type of changes

- Mostly CI/CD automation, contribution experience

## Checklist

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,18 @@
+---
+name: CI
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+  workflow_dispatch:
+
+jobs:
+  statics:
+    name: Static checks
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+      - uses: pre-commit/action@v1.0.1
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -5,7 +5,4 @@ balanced_wrapping = true
 combine_as_imports = true
 case_sensitive = true
 
-known_first_party = kopf
-
-filter_files = true
 skip_glob = examples/*.py
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,16 +12,6 @@ branches:
   - /^release\/.*$/
   - /^\d+\.\d+(\.\d+)?(\.?(rc|dev|pre|post)\d+)?(-\S*)?(\+\S*)?$/
 
-# Linting is super-fast and lightweight, gives the near-instant feedback if code is wrong.
-# It does not depend on the environment (services, dependencies, even the framework itself).
-# Currently, most linters are optional and performed manually from time to time. This will change.
-_linting: &linting
-#  install: []  # no requirements.txt  # TODO: Uncomment again when pre-commit is used (saves ~15s)
-  services: []  # no docker & co
-  before_script: []
-  script:
-    - isort . --check --diff || true
-    - isort examples --settings=examples --check --diff || true
 
 # Unit-tests depend on a lightweight environment of the framework and its dependencies only, no K8s.
 # Type-checking also needs the dependencies installed, but it goes first for faster failure.
@@ -29,7 +19,6 @@ _unittests: &unittests
   services: []  # no docker & co
   before_script: []
   script:
-    - mypy kopf --strict --pretty
     - pytest --cov=kopf --cov-branch
   after_success:
     - coveralls || true
@@ -88,10 +77,8 @@ jobs:
   - { <<: *realistic, env: K3S=v1.16.14+k3s1 }
   - { <<: *realistic, env: K3S=v1.16.14+k3s1 CRDAPI=v1beta1 }
 
-  # Linting is super-fast and lightweight (40s), gives the near-instant feedback.
   # Unit-tests are relatively fast (2m), but no integration tests, no k8s installed.
   # This pack of jobs takes roughly 5m, which is a duration of one realistic job.
-  - { <<: *linting, name: Linting and static analysis }
   - { <<: *unittests, python: "3.7" }
   - { <<: *unittests, python: "3.8" }
 
--- a/examples/.isort.cfg
+++ b/examples/.isort.cfg
@@ -4,8 +4,3 @@ multi_line_output = 11
 balanced_wrapping = true
 combine_as_imports = true
 case_sensitive = true
-
-known_third_party = kopf
-
-filter_files = true
-skip_glob = kopf/**
--- a/kopf/config.py
+++ b/kopf/config.py
@@ -7,7 +7,6 @@ LOGLEVEL_WARNING = logging.WARNING
 LOGLEVEL_ERROR = logging.ERROR
 LOGLEVEL_CRITICAL = logging.CRITICAL
 
-
 # DEPRECATED: Used for initial defaults for per-operator settings (see kopf.structs.configuration).
 class EventsConfig:
     """
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,6 @@ env
 
 # VSCode
 .vscode
+
+# InteliJ IDEs
+.idea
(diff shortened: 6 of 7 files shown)


turbaszek:
I would like to propose two things:
- introducing pre-commits to ease checking code quality, it's a pretty powerful framework and helps a lot and allows developers to focus on code
- start using Github Actions - those have better integration with Github and are more robust

@nolar what do you think?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,18 @@
...
+  workflow_dispatch:
+
+jobs:
+  statics:
+    name: Static checks
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+      - uses: pre-commit/action@v1.0.1
```
nolar, line 18: Thanks for this proposal! Are there advantages in introducing GitHub Actions into play, while Travis CI is used? This would spread the CI activity to two systems, which I see as a problem.

.travis.yml
```diff
@@ -12,24 +12,13 @@ branches:
...
-    - isort . --check --diff || true
-    - isort examples --settings=examples --check --diff || true
 
 # Unit-tests depend on a lightweight environment of the framework and its dependencies only, no K8s.
 # Type-checking also needs the dependencies installed, but it goes first for faster failure.
 _unittests: &unittests
   services: []  # no docker & co
   before_script: []
   script:
-    - mypy kopf --strict --pretty
```
nolar, line 32: I would recommend to keep it here for two reasons.

As I learned the hard way, different Python versions have different typing/typesheds for StdLib, and sometimes the type-checks succeed for one version and fail for another (cannot find the examples now though). So, each Python version (3.7, 3.8, 3.9, so on) should be type-checked on its own. That was the first reason to keep them here.

The second reason to have them with the unit-tests is that the type-check requires Kopf and its dependencies installed (to check against the dependencies' types). For static checks of pre-commit & alike, installing the library and dependencies is a waste of time, they work on source files as text, not as code. And so they can be sped up by not installing the dependencies (≈20s). But then, they will not be able to type-check. For unit-tests, installing the dependencies is unavoidable.

The type-check is usually fast, a few seconds (4.75s), much less than the unit-tests (>60s), so it is not a problem.

Here is an example where I got the numbers from:

* https://travis-ci.org/github/nolar/kopf/jobs/729061179

image

examples/.isort.cfg
```diff
@@ -4,8 +4,3 @@ multi_line_output = 11
 balanced_wrapping = true
 combine_as_imports = true
 case_sensitive = true
-
-known_third_party = kopf
-
-filter_files = true
-skip_glob = kopf/**
```
nolar, line 11: Can you please clarify what was the reason for removing these settings? Currently, they are intentionally formatted differently — with examples considered as 3rd-party apps, as if they are written not in Kopf's repo.

.isort.cfg
```diff
@@ -5,7 +5,4 @@ balanced_wrapping = true
 combine_as_imports = true
 case_sensitive = true
 
-known_first_party = kopf
-
-filter_files = true
```
nolar, line 10: I remember that a couple weeks ago, `isort` was _sometimes_ treating Kopf itself as a third-party library. I didn't get when and why. But the explicit declaration did help.

So, this is better to keep it here.

kopf/config.py
```diff
@@ -7,7 +7,6 @@
 LOGLEVEL_ERROR = logging.ERROR
 LOGLEVEL_CRITICAL = logging.CRITICAL
 
-
```
nolar, line 10: As far as I remember the styling guides and linters, there should be 2 empty lines before functions and classes in the module (but 1 empty line before constants and values of different kinds, including types (unless they use `class`)).

It would not be a blocker for this PR per se. But since my action next time I revisit this file would be to re-add the line, I would like to clarify the reason for this removal: was it intentional? Or just accidentally happened, and it is fine to re-add it?

nolar (changes requested):

nolar:
_And the overall comment that I forgot to add:_

Before merging this PR, I would like to play with GitHub Actions myself, in a separate repo, to get some hands-on experience. I never used them before, and would be careful to just jump into uncharted waters.

But this would delay the merge by a couple of weeks (depends).

.github/workflows/ci.yaml
turbaszek: I would see this as a temporary step to avoid to vast scope for single PR. 

From my experience, I would suggest migrating to Github Actions. There are multiple reasons:
- stability
- better way of defining workflows (for example using matrices) 
- better integration with Github

We experience too much pain with Travis in Apache Airflow. And since Github Actions are getting a lot of love from Github I would say that this is a better way forward 😉

kopf/config.py
turbaszek: In case of formatting I would love to propose using black. Then there's no need to worry about anything. But that's something out of the scope of this PR as it will introduce a lot of changes and should be discussed before.

turbaszek:
> Before merging this PR, I would like to play with GitHub Actions myself, in a separate repo, to get some hands-on experience. I never used them before, and would be careful to just jump into uncharted waters.

Happy to help, we use them heavily in Apache Airflow (and other projects as well) 😉 
https://github.com/apache/airflow/blob/master/.github/workflows/ci.yml

.github/workflows/ci.yaml
nolar: I see. Thanks for clarifing.

From my point of view, Travis has stability, matrices, and sufficiently good integration with GitHub. I had no problems with it so far. It would be good if you can share an article or a talk what kind of problem you had with it — maybe it would prevent Kopf from stepping into the same problems.

For now, sorry, I cannot take this as an objective argument for the switch. I take this as your opinion based on your experience and your preferences. Sadly, I cannot form my own opinion right now — it would require time (see another comment — to play with GitHub Actions myself in a separate repo).

For this PR, I suggest to keep Travis CI as the one and the only CI solution in the project. When and if the migration to GitHub Action happens after a thorough consideration, this should be done consistently for all build steps (static analysis, unit-tests, python versions, integration tests, k8s versions (k3s/k3d), lengthy nightly tests (minikube), etc). Consistency has value.

I've created #554 to consider GitHub Actions (and I see one big problem there already — free minutes limit).

---

Regarding `pre-commit` itself: I like it, I am all for it, but with common sense applied: the little issues in styling/formatting should not prevent the contributors to contribute. 

It is fine if the codebase is not perfectly styled (again, within common sense). The project is not that popular to have thousands of developers fighting over the code-style, so that it could sacrifice some percentage of them. 

I abandoned my attempts to contribute to some good projects because the contribution experience turned into fighting with spaces and commas of their style checkers. Or worse, with code uglifiers like Black (I will be strongly against that kind of things).

I don't want people to feel the same here. Every contribution matters! (but again: common sense).

turbaszek: > I abandoned my attempts to contribute to some good projects because the contribution experience turned into fighting with spaces and commas of their style checkers. Or worse, with code uglifiers like Black (I will be strongly against that kind of things).

From the maintainer point of view - code quality is as important as number of contributors. The more is done automatically (formatting, linting, checking etc) the more attention maintainer can pay to review the changes. Avoiding using mainstream and well-known approaches (black, mypy, pylint) results in code that is not consistent, thus hard to read. But I fully understand you, it's your project.

> I don't want people to feel the same here. Every contribution matters! (but again: common sense).

I think what matters is how much help people get from maintainers. In Airflow we had 40+ checks including, linting and checking some wording and missing tests. And still, we see a lot of people who answer to good first issues and are surprised by the ease of contributing.

turbaszek:
I'm closing this PR. May revisit it once I got more time for discussion and adjusting

.github/workflows/ci.yaml
nolar: That’s exactly my point: a project with 7k forks and 50 PRs/week has slightly different rules of play, requirements, and solutions. 

When and if Kopf will be at these levels, I will be actively advertising for Black myself — because it makes sense and has reasons: the disadvantages of not using it outweigh the disadvantages of using it. 

Until then, let’s keep it simple.

PR #562, opened 2020-10-07
Fix generics & MROs for Python 3.9 and Kopf 0.27

nolar:
## What do these changes do?

Enable Kopf 0.27.x for Python 3.9 by fixing an issue when some of the classes were reporting `TypeError: Cannot create a consistent method resolution order (MRO) for bases Generic, ...` on their declaration (not at runtime).

Since nothing changes essentially in the codebase, I assume the change is safe enough. That was surprisingly easy (easier than I thought it will be).

This PR is the same as the primary one — https://github.com/nolar/kopf/pull/561 — but for 0.27 (to be released as 0.27.1).

Related: https://github.com/k8spin/k8spin-operator/pull/24


nolar:
0.27.1 release has failed: the PyPI credentials are changed too. So as the whole CI pipeline. That codebase is way too old.

nolar:
And fixed with #563

PR #565, opened 2020-10-08
split logging to stdout and stderr

mareckii:
## What do these changes do?
Split logging severity between stdout / stderr
INFO, DEBUG - stdout
WARN,CRITICAL  -  stderr 

## Description

All logs were logged to stderr

INFO, DEBUG - stdout
WARN,CRITICAL  -  stderr 

Yes

## Type of changes

## Checklist

- [X ] The code addresses only the mentioned problem, and this problem only
- [X ] I think the code is well written

### DIFF
--- a/kopf/engines/loggers.py
+++ b/kopf/engines/loggers.py
@@ -12,6 +12,7 @@ import asyncio
 import copy
 import enum
 import logging
+import sys
 from typing import Any, MutableMapping, Optional, Tuple
 
 import pythonjsonlogger.jsonlogger
@@ -207,10 +208,16 @@ def configure(
 ) -> None:
     log_level = 'DEBUG' if debug or verbose else 'WARNING' if quiet else 'INFO'
     formatter = make_formatter(log_format=log_format, log_prefix=log_prefix, log_refkey=log_refkey)
-    handler = logging.StreamHandler()
-    handler.setFormatter(formatter)
+    handler_stdout = logging.StreamHandler(sys.stdout)
+    handler_stdout.setFormatter(formatter)
+    handler_stdout.addFilter(lambda record: record.levelno <= logging.INFO)
+    handler_stdout.setLevel(logging.DEBUG)
+    handler_stderr = logging.StreamHandler(sys.stderr)
+    handler_stderr.setFormatter(formatter)
+    handler_stderr.setLevel(logging.WARNING)
     logger = logging.getLogger()
-    logger.addHandler(handler)
+    logger.addHandler(handler_stdout)
+    logger.addHandler(handler_stderr)
     logger.setLevel(log_level)
 
     # Prevent the low-level logging unless in the debug mode. Keep only the operator's messages.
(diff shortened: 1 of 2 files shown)


nolar:
@mareckii Thanks for your contribution.

Initially, I had some mixed feelings about that, so I decided to take several days to consider it carefully, with all pros & cons.

I see how it can be beneficial in some environments — e.g. PyCharm — where the stdout & stderr are coloured differently: regular white/black for stdout, highlighted red for stderr — which makes the error detection easier.

However, there is one significant problem hidden in this change: stdout and stderr are often consumed out of sync by log agents and log readers. Without the split, the messages are logged to the same stream in the same order they were issued. With this change, they can end up in the wrong order — with causes of an error going after the error is logged, or just lost due to terminated pod.

This might damage the experience for the majority of users in the K8s-deployed environments for the benefit of less often scenarios with local development and stdout/stderr colouring.

For this reason, I decided not to merge this PR. I'm very sorry about that.

Please note, that I have to reject only the log separation to stdout+stderr, not logging to stdout per se. If you wish, you can implement a CLI flag `--log-stdout` to use stdout for all log messages instead of the current default stderr. The starting point for that would be in [`cli.py`](https://github.com/nolar/kopf/blob/0.28/kopf/cli.py#L36-L41).

---

If you look for alternative solutions without changing the framework, this same effect can be achieved by one of the following ways:

**Option A.** Configure the logging in your specific way in a startup handler of your specific operator (see [Startup](https://kopf.readthedocs.io/en/latest/startup/), [Configuration](https://kopf.readthedocs.io/en/latest/configuration/)):

```python
import kopf, logging

@kopf.on.startup()
def configure(**_):
    handler_stdout = ...
    handler_stderr = ...
    logger = logging.getLogger()
    logger.handlers[:] = [handler_stdout, handler_stderr]
```

If running in `--verbose` mode, one or two "DEBUG" lines could be logged before the handler is invoked and the logging is reconfigured. To solve that, the abovementioned `--log-stdout` can be used (after implemented) to log to stdout by default.

**Option B.** Implement your own CLI for your operator, maybe without any CLI options at all (if the operator is specific enough), configure the logging your own way (e.g. as in this PR), and then run an embedded operator. It is a matter of a dozen lines of code. See the documentation on how the operators can be embedded into scripts: 

* https://kopf.readthedocs.io/en/latest/embedding/

Synchronous [`kopf.run()`](https://kopf.readthedocs.io/en/latest/packages/kopf/?highlight=kopf.operator#kopf.run) & asynchronous [`kopf.operator()`](https://kopf.readthedocs.io/en/latest/packages/kopf/?highlight=kopf.operator#kopf.operator) are part of the public interface, their kwargs are too, so they will be maintained backward-compatible with the new releases.

mareckii:
@nolar 
Thanks for your answer. I already implemented loggers in may app.
Just want to add that this is not only to have better visibility on PyCharm, using gke + fluentd + stackdriver (cloud logs) for logging everything on stderr is flagged as error. This was root cause of my change.

PR #570, opened 2020-10-26
docs: add `pykorm` library

Frankkkkk:
Hi @nolar,

This is just a small documentation change to promote my project [Pykorm](https://github.com/Frankkkkk/pykorm) which is a python kubernetes ORM. I use it a lot with kopf and I think that it could be useful for some of kopf's users.

Cheers  and have a nice week,
Frank


nolar:
Okay 😄 

Though, please do not expect that this whole section/"note" will remain there forever. For a few months — for sure. I have some pending thoughts on refactoring the docs massively.

Issue #571, opened 2020-10-29
One handler function attached to 2 events causes 2 fields in status to be created instead of one?


Jc2k:
## Long story short

If i decorate a function with 2 different `@kopf.on` handlers they seem to update different fields in `status`.
 This means I can't have a `kopf.on.create` and `kopf.on.field` both update `status.some_field`, i'll actually end up with 2 fields instead.

I`m pretty sure this didn't use to happen, but i don't know if i "got lucky" before...

## Description

Recently upgraded an older kopf based operator. It had something like this:

```python
@kopf.on.create("foo.example.com", "v1", "bar")
@kopf.on.field("foo.example.com", "v1", "bar", field="spec.some_field")
def my_config_map(*, body, logger, **kwargs):
    return "some-resource-name-12345-abc123c"
```

It creates a resource (lets say a ConfigMap for the sake of argument).

Previously the status was set to:

```json
{
  "status": {
    "my_config_map": "some-resource-name-12345-abc123c",
  }
}
```

(0.23ish)

But now im seeing something like

```json
{
  "status": {
    "my_config_map": "some-resource-name-12345-abc123d",
    "my_config_map/spec.some_field": "some-resource-name-12345-abc123c",
  }
}
```

With 0.28.

I am not sure if this is a regression or i was just getting lucky before, so on the off chance its a regression I thought i should report it. If it's not a regression is there anyway to force them both to write to `my_config_map`? This PR looks like it would give me enough control to that, but otherwise i'm not sure how to proceed. https://github.com/zalando-incubator/kopf/pull/320

## Environment

* Kopf version: 0.28
* Kubernetes version: 1.19.2
* Python version: 3.8
* OS/platform: linux

nolar:
Hello.

The described behaviour seems to be as expected. The handlers write their functions' results into `.status.{handler_id}`. These two handlers would always have different handler ids, one should always be suffixed with the field name. I see no reason why they should be putting the result into one field even in the past versions of Kopf.

And it is impossible to assign explicit ids to both handlers via `id='...'` option to the decorators. Though it is possible, but I am not sure what is the expected outcome and what are the side-effects of that — they will be conflicting for their own state.

One reason why you were lucky in the past, most likely, is that on-field handlers were an equivalent of on-update limited to a field context for some time. Later (in 0.28), they were fixed to be triggered really in the context of any field change — including the field's creation (a change from `None` to a value), both when the field is added and when the whole object is added.

[...]

Issue #566, opened 2020-10-09
Field handler is trigger even when resource is first created


euan-tilley:
## Long story short

While #375 appears to have been fixed the field handler is still being triggered when a resource is first created. This didn't happen in 0.26.

## Description

<details><summary>The code snippet to reproduce the issue</summary>

```python
import kopf

@kopf.on.create('company.com', 'v1', 'testresources')
def create_handler(spec, **_):
    pass

@kopf.on.field('company.com', 'v1', 'testresources', field='spec.createField')
def createField_changed(old, new, diff, logger, **_):
    logger.info("Diff - {}".format(diff))
```
</details>

<details><summary>The exact command to reproduce the issue</summary>

```bash
kopf run --standalone test-handler.py

cat <<EOF | kubectl create -f -
apiVersion: company.com/v1
kind: TestResource
metadata:
  name: test
spec:
  createField: true
EOF
```
</details>

<details><summary>The full output of the command that failed</summary>

```
[2020-10-09 18:14:43,256] kopf.activities.auth [INFO    ] Activity 'login_via_client' succeeded.
[2020-10-09 18:14:43,256] kopf.reactor.activit [INFO    ] Initial authentication has finished.
[2020-10-09 18:14:48,484] kopf.objects         [INFO    ] [default/test] Handler 'create_handler' succeeded.
[2020-10-09 18:14:48,600] kopf.objects         [INFO    ] [default/test] Diff - (('add', (), None, True),)
[2020-10-09 18:14:48,601] kopf.objects         [INFO    ] [default/test] Handler 'createField_changed/spec.createField' succeeded.
[2020-10-09 18:14:48,601] kopf.objects         [INFO    ] [default/test] Creation event is processed: 2 succeeded; 0 failed.
```
</details>

## Environment

* Kopf version: 0.28
* Kubernetes version: minikube v0.35.0 on darwin (amd64) & v.1.11.9
* Python version: 3.8.0
* OS/platform: macos & coreos

<details><summary>Python packages installed</summary>

```
aiohttp==3.6.2
aiojobs==0.2.2
async-timeout==3.0.1
attrs==19.3.0
cachetools==4.1.0
certifi==2020.4.5.2
cffi==1.14.0
chardet==3.0.4
click==7.1.2
cryptography==2.9.2
google-auth==1.17.2
idna==2.9
iso8601==0.1.12
Jinja2==2.11.2
kopf==0.28
kubernetes==11.0.0
MarkupSafe==1.1.1
[...]

PR #573, opened 2020-11-02
Field filters for all resource handlers

nolar:
Implement field & value filters for all resource-changing handlers: `@kopf.on.resume/create/update/delete`.

_It was a long-needed feature to replace field handlers (`@kopf.on.field`) with more clear, explicit, and less ambiguous syntax. It was drafted in Jan'2020, but didn't get into the main codebase due to some semantical complications. Over the past year, the complications were resolved, so it seems easy to implement this feature now._

If `field=` is set on any handler, the handler is triggered only if the field is present in the object, and `old`/`new`/`diff` kwargs are reduced to that field instead of the whole body; the usually `body` & co kwargs still point to the full body, as expected.

```python
@kopf.on.create(..., field='spec.field')
def fn(new, **_):
  print(f"{new!r}")
```

As with other filters (labels, annotations), either specific values are accepted and checked for equality, or `kopf.ABSENT`/`kopf.PRESENT` markers are accepted, or callbacks that accept a positional value and all typical kwargs, and return a boolean for whether it matches or not.

```python
@kopf.on.create(..., field='spec.field', value='value')
@kopf.on.create(..., field='spec.field', value=kopf.PRESENT)
@kopf.on.create(..., field='spec.field', value=kopf.ABSENT)
@kopf.on.create(..., field='spec.field', value=lambda val, **_: 'a' in val)
@kopf.on.create(..., field='spec.field', value=kopf.any_([fn1, fn2]))
@kopf.on.create(..., field='spec.field', value=kopf.all_([fn1, fn2]))
@kopf.on.create(..., field='spec.field', value=kopf.none_([fn1, fn2]))
def fn(**_): pass
```

The default behaviour depends on what kind of a filter/handler it is:

For single-value filters (creation/resuming/deletion handlers), by default (i.e. `value=None`), the field must be present, and it is equivalent to `value=kopf.PRESENT`:

```python
@kopf.on.create(..., field='spec.field')
def fn(**_): pass
```

For the update & on-field handlers, additional criteria can be used: `old` & `new`. By default, it is only triggered if the value is changed (including the case when it is changed from anything to `None` — i.e. the field is removed):

```python
@kopf.on.update(..., field='spec.field')  # if changed (incl. if removed)
@kopf.on.update(..., field='spec.field', old='x', new='y')  # only this specific change
@kopf.on.update(..., field='spec.field', old=kopf.ABSENT, new=kopf.PRESENT)  # the field was set
@kopf.on.update(..., field='spec.field', old=kopf.PRESENT, new=kopf.ABSENT)  # the field was unset
def fn(**_): pass
```

Using both `value=` & `old=/new=` is prohibited, as it is confusing. It is either-either. Essentially, `new` is the same as `value`.

---

Some unusual tricks are now possible — as side-effects, not as the main goal of this change:

Trigger a handler only if the required field is NOT defined — the opposite of the default behaviour. The value of `new` will be `None`:

```python
@kopf.on.create(..., field='spec.nofield', value=kopf.ABSENT)
def fn(**_): pass
```

---

`@kopf.on.field` handler are now discouraged — but not yet deprecated. They will most likely remain for some long time for backward compatibility. A field handler like this:

```python
@kopf.on.field(..., field='spec.field')
def fn(old, new, **_):
    pass
```

Is equivalent to this:

```python
@kopf.on.create(..., field='spec.field')
@kopf.on.update(..., field='spec.field')
def fn(old, new, **_):
    pass
```

Note: but not for deletion & resuming, as nothing changes in those occasions, and on-field is only triggered when something is changed. 

---

The field-value (but not old/new) filtering logic can also be applied to all other resource handlers:

```python
import kopf

@kopf.on.event(..., field='spec.field', value='value')
def event_fn(**_):
    pass

@kopf.timer(..., field='spec.field', value='value', interval=1)
def timer_fn(**_):
    pass

@kopf.daemon(..., field='spec.field', value='value')
def event_fn(stopped, **_):
    stopped.wait(1000)
```

Their semantics is the same as the labels/annotations/callback filtering: the handler is invoked and the daemon is running as long as the criterion is satisfied. Once the resource changes so that it does not satisfy the criterion, the handlers are not invoked, and the daemon is stopped.

---

Remaining TODOs:

* [x] **Test things manually and verify that the behaviour is consistent and intuitive.**
* [x] Raise field&value (but not old&new) filters to all resource handlers: on.event, daemon, timer.
* [x] Performance optimisations for dict lookups.
* [x] Unit-tests.
* [x] An example (or extend the example 05-handlers?).
* [x] Documentation (preview: https://kopf.readthedocs.io/en/field-filters/filters/)

Related: #571 #566

### DIFF
--- a/docs/filters.rst
+++ b/docs/filters.rst
@@ -2,78 +2,76 @@
 Filtering
 =========
 
-.. highlight:: python
+Handlers can be restricted to only the resources that match certain criteria.
 
-It is possible to only execute handlers when the object that triggers a handler
-matches certain criteria.
+Multiple criteria are joined with AND, i.e. they all must be satisfied.
 
-The following filters are available for all resource-related handlers
-(event-watching and change-detecting):
+Unless stated otherwise, the described filters are available for all handlers:
+resuming, creation, deletion, updating, event-watching, timers, daemons,
+or even to sub-handlers (thus eliminating some checks in its parent's code).
 
+There are only a few kinds of checks:
 
-By labels
-=========
+* Specific values -- expressed with Python literals such as ``"a string"``.
+* Presence of values -- with special markers ``kopf.PRESENT/kopf.ABSENT``.
+* Per-value callbacks -- with anything callable and evaluatable to true/false.
+* Whole-body callbacks -- with anything callable and evaluatable to true/false.
 
-* Match an object's label and value::
-
-    @kopf.on.create('zalando.org', 'v1', 'kopfexamples',
-                    labels={'some-label': 'somevalue'})
-    def my_handler(spec, **_):
-        pass
-
-* Match on the existence of an object's label::
-
-    @kopf.on.create('zalando.org', 'v1', 'kopfexamples',
-                    labels={'some-label': kopf.PRESENT})
-    def my_handler(spec, **_):
-        pass
-
-* Match on the absence of an object's label::
-
-    @kopf.on.create('zalando.org', 'v1', 'kopfexamples',
-                    labels={'some-label': kopf.ABSENT})
-    def my_handler(spec, **_):
-        pass
+But there are multiple places where these checks can be applied,
+each has its own specifics.
 
 
-By annotations
-==============
+Metadata filters
+================
 
-* Match on object's annotation and value::
+Metadata is the most commonly filtered aspect of the resources.
+
+Match only when the resource's label or annotation has a specific value:
+
+.. code-block:: python
 
     @kopf.on.create('zalando.org', 'v1', 'kopfexamples',
+                    labels={'some-label': 'somevalue'},
                     annotations={'some-annotation': 'somevalue'})
     def my_handler(spec, **_):
         pass
 
-* Match on the existence of an object's annotation::
+Match only when the resource has a label or an annotation with any value:
+
+.. code-block:: python
 
     @kopf.on.create('zalando.org', 'v1', 'kopfexamples',
+                    labels={'some-label': kopf.PRESENT},
                     annotations={'some-annotation': kopf.PRESENT})
     def my_handler(spec, **_):
         pass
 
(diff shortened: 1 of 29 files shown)


nolar:
Documentation preview: https://kopf.readthedocs.io/en/field-filters/filters/

nolar:
Tested manually. The field-filters work as expected, it seems. Even for sub-handlers with this setup (impractical, but possible):

```python
import kopf


async def created_with_field(**kwargs):
    print(f'CREATED')
    for i in range(3):

        @kopf.on.this(field='spec.field', value='value', id=f'a{i}')
        def a_sub(new, i=i, **kwargs):
            print(f'A-SUB {i} with field: {new}')

        @kopf.on.this(field='spec.field', value=kopf.PRESENT, id=f'b{i}')
        def b_sub(new, i=i, **kwargs):
            print(f'B-SUB {i} with field: {new}')


@kopf.on.update('zalando.org', 'v1', 'kopfexamples')
async def update_with_field(**kwargs):
    print(f'UPDATED')
    for i in range(3):

        @kopf.on.this(field='spec.field', old=i-1, new=i, id=f'c{i}')
        def c_sub(new, i=i, **kwargs):
            print(f'C-SUB {i} with field: {new}')
```

And then doing:

```shell
kubectl apply -f examples/obj.yaml
kubectl patch -f examples/obj.yaml --type merge -p '{"spec": {"field": -1}}'
kubectl patch -f examples/obj.yaml --type merge -p '{"spec": {"field": 0}}'
kubectl patch -f examples/obj.yaml --type merge -p '{"spec": {"field": 1}}'
kubectl patch -f examples/obj.yaml --type merge -p '{"spec": {"field": 2}}'
kubectl patch -f examples/obj.yaml --type merge -p '{"spec": {"field": 0}}'  # no reaction
```

PR #579, opened 2020-11-12
Optionally remove Status handler_id prefix.

samgiles:
## What do these changes do?

This ports a PR that was originally raised (https://github.com/zalando-incubator/kopf/pull/320) on the Zalando repository.
Thanks @lack!

Adding an optional parameter to the decorators for resource change handlers that allows the return value to be stored directly onto the `status` key on the resource in question.

## Description

Implements the proposal in issue https://github.com/nolar/kopf/issues/319.

## Issues/PRs

https://github.com/nolar/kopf/issues/319

## Type of changes

## Checklist

- [x] Documentation reflects the changes

I can't take credit for the code itself since it was already written here: https://github.com/zalando-incubator/kopf/pull/320

I'm not sure if we've covered all the cases here, would be good to know if I've missed something.

### DIFF
--- a/docs/status.rst
+++ b/docs/status.rst
@@ -1,0 +1,59 @@
+=======
+Updating Status
+=======
+
+Kopf will take the return value of all state-changing handlers (even
+sub-handlers) and add them to the ``Status`` of the corresponding Kuberenetes
+resource.  By default, this is stored under the handler id which by default is 
+the function name.
+
+.. note::
+
+    The ``CustomResourceDefinition`` requires that the fields that will be set in
+    status have a corresponding schema.
+
+    You could also use ``x-kubernetes-preserve-unknown-fields: true``::
+
+      schema:
+        openAPIV3Schema:
+          type: object
+          properties:
+            status:
+              type: object
+              x-kubernetes-preserve-unknown-fields: true
+            spec:
+              ...
+
+Given the following handler definition::
+
+    import kopf
+
+    @kopf.on.create('zalando.org', 'v1', 'kopfexamples')
+    def my_handler(spec, **_):
+        return {'field': 'value}
+
+The resulting object will have the following data::
+
+    spec:
+      ...
+    status:
+      my_handler:
+        field: value
+
+In order to remove the handler ID from the result, you may set the optional
+``status_prefix=False`` when defining the handler.
+
+So with the handler definition::
+
+    import kopf
+
+    @kopf.on.create('zalando.org', 'v1', 'kopfexamples', status_prefix=False)
+    def my_handler(spec, **_):
+        return {'field': 'value}
+
+The resulting object will have the following data::
+
+    spec:
+      ...
+    status:
+      field: value
--- a/kopf/on.py
+++ b/kopf/on.py
@@ -145,6 +145,7 @@ def resume(  # lgtm[py/similar-function]
         labels: Optional[filters.MetaFilter] = None,
         annotations: Optional[filters.MetaFilter] = None,
         when: Optional[callbacks.WhenFilterFn] = None,
+        status_prefix: bool = True,
 ) -> ResourceChangingDecorator:
     """ ``@kopf.on.resume()`` handler for the object resuming on operator (re)start. """
     def decorator(  # lgtm[py/similar-function]
@@ -161,6 +162,7 @@ def resume(  # lgtm[py/similar-function]
             labels=labels, annotations=annotations, when=when,
             initial=True, deleted=deleted, requires_finalizer=None,
             reason=None,
+            status_prefix=status_prefix,
         )
         real_registry.resource_changing_handlers[real_resource].append(handler)
         return fn
@@ -180,6 +182,7 @@ def create(  # lgtm[py/similar-function]
         labels: Optional[filters.MetaFilter] = None,
         annotations: Optional[filters.MetaFilter] = None,
         when: Optional[callbacks.WhenFilterFn] = None,
+        status_prefix: bool = True,
 ) -> ResourceChangingDecorator:
     """ ``@kopf.on.create()`` handler for the object creation. """
     def decorator(  # lgtm[py/similar-function]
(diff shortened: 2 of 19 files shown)


docs/status.rst
```diff
@@ -0,0 +1,59 @@
...
+    @kopf.on.create('zalando.org', 'v1', 'kopfexamples', status_prefix=False)
+    def my_handler(spec, **_):
+        return {'field': 'value}
+
+The resulting object will have the following data::
+
+    spec:
+      ...
+    status:
+      field: value
```
nolar, line 59: **Important:** What will happen with `return "hello"` or `return 123` or other scalars?  What about `return ["abc", "def"]`?

docs/status.rst
```diff
@@ -0,0 +1,59 @@
+=======
+Updating Status
+=======
```
nolar, line 3:
```suggestion
===============
Updating Status
===============
```

docs/status.rst
```diff
@@ -0,0 +1,59 @@
...
+
+Kopf will take the return value of all state-changing handlers (even
+sub-handlers) and add them to the ``Status`` of the corresponding Kuberenetes
```
nolar, line 6:
```suggestion
sub-handlers) and add them to the ``.status`` stanza of the corresponding Kubernetes
```

docs/status.rst
```diff
@@ -0,0 +1,59 @@
...
+            spec:
+              ...
+
+Given the following handler definition::
+
+    import kopf
+
+    @kopf.on.create('zalando.org', 'v1', 'kopfexamples')
+    def my_handler(spec, **_):
+        return {'field': 'value}
```
nolar, line 33: Can you please wrap all code blocks here (before and after this comment) to their specific language highlights?

This is one of the following instead of `::`:

```
.. code-block:: yaml
.. code-block:: python
```

I'm highly uncertain if there is any kind of automatic language detection in SphinxDocs.

kopf/storage/states.py
```diff
@@ -304,9 +305,16 @@ def deliver_results(
...
-            patch.setdefault('status', {}).setdefault(handler_id, {}).update(outcome.result)
+            if hasattr(outcome.handler, 'status_prefix') and not outcome.handler.status_prefix:
+                base = patch.setdefault('status', {})
+            else:
+                base = patch.setdefault('status', {}).setdefault(handler_id, {})
+            base.update(outcome.result)
         else:
-            patch.setdefault('status', {})[handler_id] = copy.deepcopy(outcome.result)
+            if hasattr(outcome.handler, 'status_prefix') and not outcome.handler.status_prefix:
+                patch.setdefault('status', {})[copy.deepcopy(outcome.result)] = {}
```
nolar, line 315: **Important:** I'm quite not sure what does this line mean. Can you please clarify the intentions? It looks like the result is used as the key, not as the value.

kopf/on.py
```diff
@@ -145,6 +145,7 @@ def resume(  # lgtm[py/similar-function]
         labels: Optional[filters.MetaFilter] = None,
         annotations: Optional[filters.MetaFilter] = None,
         when: Optional[callbacks.WhenFilterFn] = None,
+        status_prefix: bool = True,
```
nolar, line 148: **Important:** Generally, as a rule of thumb, all handler options are designed so that the default value is `None` and evaluates to falseness. Even for booleans (see `optional` or `deleted` in some handlers nearby — it is `Optional[bool]`). I would prefer that all user-facing DSL features follow this convention in the future.

Can you please negate the option and its meaning? (I am not sure what would be the better wording to describe this: flat status? direct status? raw status? result_to_status? something else?)

nolar (changes requested):

nolar:
On a side note, I thought on a similar feature in relation to a completely different purpose (cross-resource handlers). One idea that I had is to separate the result delivery from the handler declaration completely — via `@kopf.to....` decorators of different kind. Your goal could be achieved with:

```python
@kopf.on.resume('zalando.org', 'v1', 'kopfexamples')
@kopf.to.status(id='')
def fn(**_):
    return {'hello': 'world'}
```

The absence of `@kopf.to` decorators, for backward compatibility, would be equivalent to `@kopf.to.status()`, with `id` equal to the function name — same as the `id` option for the handlers now.

Why so? In the long-term, this would allow to extend it with multiple delivery targets, including those not even in this same resource:

```python
@kopf.on.resume('zalando.org', 'v1', 'kopfexamples')
@kopf.to.status(id='')
@kopf.to.owners(field='status.children.{}')  # via ownerReferences, to all of them if there are many
@kopf.to.owners(kind='KopfAnotherExample')  # via ownerReferences, to only a specific kind
@kopf.to.related(kind='KopfAnotherExample', via=...)  # matching criteria are uncertain here (yet)
def fn(**_):
    return {'hello': 'world'}
```

Usage of decorators allows flexibility with some options for each delivery target, and multiple targets at once.

The exact semantics is not yet fully clarified, so consider the snippets above just as ideas.

Implementation-wide, I thought about storing the "to" values directly on a function as a hidden attribute (e.g. `_kopf_to_targets=[]` as a list of these target definitions), so that all handlers (create/resume in this example) would deliver to all declared targets. It means that the target is not a property of the handler, but of the function.

While all these types of targets are not in the nearest scope for implementation, is there a way to consider using `@kopf.to.status()` in this implementation? How much more difficult would it be?

It is not a problem if this is not implemented now. There is a clear visible way of ensuring backward compatibility with the handlers' option when it is implemented (with some nuances for double-handling functions (e.g. create+resum) with different setup). So, if the complexity is too high, it is better to postpone this solution for later, and implement it as it is done in this PR via handlers' properties.

caffeinism:
any plans now?

nolar:
On the `@kopf.to.` syntax — no, there are no current plans, the priorities are different.

On this PR — I will close it for now: there are too many conflicts, they are difficult to resolve, and some issues from the comments are not solved (and I'm not sure how to solve them properly on the user-story level: e.g., `return 123`).

As a workaround to achieve the desired effect, use the `patch` kwarg and populate it as needed instead of `return`'ing values from the handler (the snippet from #319):

```python
import kopf

@kopf.on.create("zalando.org", "v1", "kopfexamples")
async def created(spec, patch, **kwargs):
    patch.status["result"] = "Created"  # instead of: return {"result": "Created"}

@kopf.on.update("zalando.org", "v1", "kopfexamples")
async def updated(spec, patch, **kwargs):
    patch.status["result"] = "Updated"
```

With this tool in hand, adding the extra flag does not look significantly more beneficial, but adds complexity and maintenance burden.

PS: Extracted to #827.

Issue #613, opened 2020-12-14
is_namespaced is not always discovered correctly


amadev:
## Long story short

In https://github.com/nolar/kopf/blob/214310c8f678fad5e267aacfbbc8acdadb557d9c/kopf/clients/discovery.py#L17
context._discovered_resources may not contain all necessary info. For example, if Resource(group='', version='v1', plural='secrets')
is discovered first, it's filled with 'v1': {}. Thus, node resource discovered wrongly after that.

## Description

[2020-12-14 17:26:46,627] kopf.reactor.running [ERROR   ] Root task 'watcher of nodes' is failed: 404, message='Not Found', url=URL('https://172.16.244.44:6443/api/v1/namespaces/openstack/nodes')
Traceback (most recent call last):
  File "/home/amadev/m/openstack-controller/.tox/dev/lib/python3.7/site-packages/kopf/reactor/running.py", line 438, in _root_task_checker
    await coro
  File "/home/amadev/m/openstack-controller/.tox/dev/lib/python3.7/site-packages/kopf/reactor/queueing.py", line 128, in watcher
    async for raw_event in stream:
  File "/home/amadev/m/openstack-controller/.tox/dev/lib/python3.7/site-packages/kopf/clients/watching.py", line 71, in infinite_watch
    async for raw_event in stream:
  File "/home/amadev/m/openstack-controller/.tox/dev/lib/python3.7/site-packages/kopf/clients/watching.py", line 110, in streaming_watch
    async for raw_event in stream:
  File "/home/amadev/m/openstack-controller/.tox/dev/lib/python3.7/site-packages/kopf/clients/watching.py", line 128, in continuous_watch
    items, resource_version = await fetching.list_objs_rv(resource=resource, namespace=namespace)
  File "/home/amadev/m/openstack-controller/.tox/dev/lib/python3.7/site-packages/kopf/clients/auth.py", line 46, in wrapper
    return await fn(*args, **kwargs, context=context)
  File "/home/amadev/m/openstack-controller/.tox/dev/lib/python3.7/site-packages/kopf/clients/fetching.py", line 99, in list_objs_rv
    response.raise_for_status()
  File "/home/amadev/m/openstack-controller/.tox/dev/lib/python3.7/site-packages/aiohttp/client_reqrep.py", line 946, in raise_for_status
    headers=self.headers)
[...]

PR #615, opened 2020-12-15
Fix discovery: check resources of api_version exist

amadev:
## What do these changes do?

This change adds an additional check that resource exists in context._discovered_resources.

## Description

Detailed info in the issue.

## Issues/PRs

issue: https://github.com/nolar/kopf/issues/613

## Type of changes

## Checklist

- [ ] I think the code is well written

### DIFF
--- a/kopf/clients/discovery.py
+++ b/kopf/clients/discovery.py
@@ -16,7 +16,9 @@ async def discover(
     if context is None:
         raise RuntimeError("API instance is not injected by the decorator.")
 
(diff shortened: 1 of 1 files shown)


nolar:
The tests were failing for the example 10 — a known issue. It was fixed in 0.28.x in #619 (back-ported from the main branch from #608). The failure is not related to this PR, so it should be fine. I have double-checked in a follow-up PR #618 — the tests are fine with both these PRs merged.

PR #617, opened 2020-12-15
Rename @kopf.on.this → @kopf.subhandler for readability

### LINKED ISSUE #616: Add more readable @kopf.subhandler instead of @kopf.on.this/kopf.register
## Background

A long time ago, when Kopf was only prototyped, sub-handlers were implemented via `@kopf.on.this()` decorator. The logic was so that if we all other handlers as `@kopf.on.{something}`, sub-handler should be also declared to be triggered "on" something, and keep sub-handlers declarative. That naming turned to be highly confusing, and so a more imperative way was introduced: `kopf.register()` function (not a decorator). And also even more imperative way: `await kopf.execute()` function.

Much later, daemons and timers were added as `@kopf.daemon` and `@kopf.timer`, because they could not be expressed with the "on-something" expression (the timers could be "on-schedule", but the daemons definitely could not be named that way).

And so, the "on-something" semantics was already violated. 

## Proposal

It makes no sense to keep it as `@kopf.on.this` anymore. Why not make it `@kopf.subhandler`, and return the good old declarative way of declaring the sub-handlers — in addition to the existing imperative ways?

## Example

```python
import kopf

@kopf.on.creation('kopfexamples')
def create_fn(name, **_):

    for i in range(5):

        @kopf.subhandler(id=str(i))
        def sub_fn(i=i, **_):
            print(f"{name}: {i}")
```

Looks more readable than `@kopf.on.this()`.

## Side-changes

Also, consider `@kopf.sub()` as a shortcut.

Also, consider adding the on-creation/on-deletion aliases for the existing on-create/on-delete decorators — for English readability.

However, the aliases go against Python Zen:

```
There should be one-- and preferably only one --obvious way to do it.
```

nolar:
Additionally, fix the docs to show that `kopf.execute()` is a coroutine, not a function.

fixes #616

### DIFF
--- a/docs/handlers.rst
+++ b/docs/handlers.rst
@@ -235,24 +235,25 @@ partials (`functools.partial`), or the inner functions in the closures:
         - item1
         - item2
 
-Sub-handlers can be implemented either imperatively::
+Sub-handlers can be implemented either imperatively
+(where it requires :doc:`asynchronous handlers <async>` and ``async/await``)::
 
     import functools
     import kopf
 
     @kopf.on.create('zalando.org', 'v1', 'kopfexamples')
-    def create_fn(spec, **_):
+    async def create_fn(spec, **_):
         fns = {}
 
         for item in spec.get('items', []):
             fns[item] = functools.partial(handle_item, item=item)
 
-       kopf.execute(fns)
+       await kopf.execute(fns=fns)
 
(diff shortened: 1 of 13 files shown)


nolar:
Wrong taget branch. Replaced by #618.

Issue #686, opened 2021-02-16
Endless touch-dummy Annotation Patching On Pod Causes High API Load


paxbit:
## Long story short

While looking into #685 I noticed weird patching behavior from kopf regarding the `touch-dummy` annotation. As soon as a pod reaches the completed state kopf constantly patches the `touch-dummy` annotation on the pod creating significant load on the API server.

## Description

The operator has a handler watching for completed pods. It is decorated like so:
```python
    @kopf.on.resume("", "v1", "pods", labels=cfg.test_pod_selector_label_dict)
    @kopf.on.field(
        "",
        "v1",
        "pods",
        id="await_pod_term",
        field="status.phase",
        labels=cfg.test_pod_selector_label_dict,
    )
```

There also is a 10sec interval timer handling unschedulable pods.
```python
    @kopf.on.timer(
        "",
        "v1",
        "pods",
        interval=cfg.watch_for_unschedulable_pod.interval,
        sharp=cfg.watch_for_unschedulable_pod.sharp,
        labels=cfg.test_pod_selector_label_dict,
        when=lambda namespace, name, status, **_: pods.get_pod_condition_or_none_from( 
            status, pods.PodCondTypeEnum.PodScheduled, pods.PodCondStatusEnum.false
        )
        is not None,
    )
```

<details>
<summary>
I have the following pod sitting in the completed state for now ~23h. (Showing only metadata as that should be the relevant part I guess)
</summary>

```yaml
apiVersion: v1
kind: Pod
metadata:
  annotations:
    kopf.zalando.org/await_pod_term.status.phase: '{"started":"2021-02-15T22:02:03.350703","stopped":"2021-02-15T22:02:03.353473","purpose":"update","retries"
:1,"success":true,"failure":false}'
    kopf.zalando.org/last-handled-configuration: |
      {"spec":{"volumes":[{"name":"test-config","configMap":{"name":"ts-job-1179-imuvqvkxczsjxvtpmau9pw","defaultMode":420}},{"name":"rasbuild","persistentVolumeClaim":{"claimName":"job-sesam-test-1179-rasbuild"}},{"name":"default-token-xlckc","secret":{"secretName":"default-token-xlckc","defaultMode":420}}],"conta
[...]

PR #728, opened 2021-03-30
Avoid dummy-patching for out-of-scope handlers (via check_done/get_delays functions)

nolar:
In #686, a complicated bug is well-described and discussed: Kopf floods the K8s API with dummy patches with no pause between them in certain initial conditions. Specifically, when there is a persisted state of a handler that didn't finish (with either success or failure) and came of out scope during the processing cycle (e.g. due to filters).

The bug was there since the purposes were introduced in #606, but was activated and made visible by #674.

Reproducible with:

```python
import kopf


@kopf.on.startup()
def configure(settings: kopf.OperatorSettings, **_):
    # Status storage must not intervene with its data:
    settings.persistence.progress_storage = kopf.AnnotationsProgressStorage()


@kopf.on.resume("configmaps", id="aaa")
@kopf.on.update("configmaps", id="bbb",
    when=lambda **_: False,  # <<< comment it, and the bug is gone.
)
def the_handler(**_):
    print("the_handler called")
```

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  annotations:
    kopf.zalando.org/last-handled-configuration: |
      {"simulated-change":123}
    kopf.zalando.org/bbb: '{"started":"2021-03-01T12:34:03.155586","purpose":"update","retries":0,"success":false,"failure":false}'
  name: bug
```

```
kubectl delete -f _issue686.yaml
kubectl create -f _issue686.yaml
kopf run _issue686.py --verbose -n default
```

The fix (this PR) replaces the state's "done"/"delays" computation logic:

* AS-IS: _all recorded handlers with the matching purpose._
* TO-BE: _all handlers currently in scope._

If a handler falls out of scope during the processing cycle, it will not be taken into account in the next "done"/"delays" computation, even it was not finished before — this was the behaviour before "purposes" were introduced in #606.

---

TODOs:

* [x] Reverify all other side-effects.
* [x] Reverify the differences of `0` vs. `None` delays in handler states (see #686's discussion).
* [x] Check if the purpose checking in "done"/"delays" is still needed and makes sense with explicitly requested handlers.
* [ ] If the logic is correct, adjust the tests.
* [ ] Consider getting rid of "purposes" at all — but they might be needed for "extras"/"counts", where the "currently in-scope handler" might be not enough, but "ever-were in-scope handlers" should be counted.

### DIFF
--- a/kopf/reactor/activities.py
+++ b/kopf/reactor/activities.py
@@ -120,7 +120,7 @@ async def run_activity(
     handlers = registry._activities.get_handlers(activity=activity)
     state = states.State.from_scratch().with_handlers(handlers)
     outcomes: MutableMapping[ids.HandlerId, states.HandlerOutcome] = {}
-    while not state.done:
+    while not state.check_done(handlers):
         current_outcomes = await handling.execute_handlers_once(
             lifecycle=lifecycle,
             settings=settings,
@@ -130,7 +130,7 @@ async def run_activity(
         )
         outcomes.update(current_outcomes)
         state = state.with_outcomes(current_outcomes)
-        await primitives.sleep_or_wait(state.delay)
+        await primitives.sleep_or_wait(state.get_delays(handlers))
 
     # Activities assume that all handlers must eventually succeed.
     # We raise from the 1st exception only: just to have something real in the tracebacks.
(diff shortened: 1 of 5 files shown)


nolar:
Replaced by #731: There, the global state remains stateful (as by its definition), remembers which handler states are "active" or "passive", and only uses the "active" states for `.done`/`.delays`. No changes in the rest of the code are needed then, only in the state classes.

PR #748, opened 2021-05-01
Clarify type annotations for daemon stopping flags

nolar:
Previously, `Union` of sync & async checkers was used in the daemon protocol. As a result, it forced the operator developers to accept both types and do proper handling of both cases — when in fact only one checker was passed according to the daemon type (sync/async).

To simplify type-checking of operators, both sync & async checkers are now downgraded to _implementation details_ (but remain exported for backwards compatibility), while only the base class is exposed for type annotations of the `stopped` kwarg. To achieve this, some "clean" (as in, "not dirty") trickery was implemented — see the docstring of the new "waiter" class.

All public interfaces should remain supported without breaking it. Even evaluating the results of `stopped.wait()` as booleans remains there: the returned value is the flag/checker itself, but it is bool-evaluable (unless `x is True`/`x is False` conditions were used).

Related #747. Fixes #746. Replaced by #757.

### DIFF
--- a/docs/daemons.rst
+++ b/docs/daemons.rst
@@ -33,7 +33,7 @@ with ``@kopf.daemon`` and make it run for a long time or forever:
             await asyncio.sleep(10)
 
     @kopf.daemon('kopfexamples')
-    def monitor_kex_sync(stopped, **kwargs):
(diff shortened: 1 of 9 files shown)


nolar:
Solved in #757

PR #758, opened 2021-05-10
Pre-create various loggers instead of creating them on-demand

nolar:


nolar:
Closed in favour of #759

Issue #767, opened 2021-05-12
kopf run does not work with click 8.x


pshchelo:
## New click breaks kopf

New `click` 8.0 has been released on May 11, 2021
With it, the latest kopf release fails to start the `kopf run`, seems like something changed in the handling of defaults for logging configuration

## Description

<details><summary>The exact command to reproduce the issue</summary>

```bash
kopf run -m <whatever handler module>
```
</details>

<details><summary>The full output of the command that failed</summary>

```
Usage: kopf run [OPTIONS] [PATHS]...                                                                                                                                                                                                                          
Try 'kopf run --help' for help.                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                              
Error: Invalid value for '--log-format': <LogFormat.FULL: '[%(asctime)s] %(name)-20.20s [%(levelname)-8.8s] %(message)s'> is not one of 'plain', 'full', 'json'. 
```
</details>


## Environment

<!-- The following commands can help:
     `kopf --version` or `pip show kopf`
     `kubectl version`
     `python --version`
-->

* Kopf version: 1.31.0
* Kubernetes version:  1.18.14
* Python version: 3.8
* OS/platform: Ubuntu 20.04

<details><summary>Python packages installed</summary>
<!-- use `pip freeze --all` -->
relevant package installed is

```
click==8.0.0
```
</details>

Immediate workaround is to add explicitly `--log-format full` to the `kopf run` invocation

pshchelo:
let's hold it for now. I get really strange results when installing in a clean docker container, in a clean virtualenv and as pip install -e . in another virtualenv
sometimes it works, sometimes it does not. sometimes old code works with click 8.x, sometimes it does not

need to investigate further. I'd really like someone else to chime in if they see the same problem

nolar:
@pshchelo I agree. Should we now restrict to click<8.0.0 until it is fixed? So that the initial journey could work; so as Kopf upgrades to new releases in other people’s apps.

pshchelo:
@nolar seems like a least intrusive immediate option at the moment

SStorm:
Can confirm this is happening and that click 8.0.0 seems incompatible. Please can we fix versions of all dependencies to avoid this happening in the future?

SStorm:
On a related note, this not only breaks the latest kopf release, but all previous kopf releases too :(

We worked around it in our app by fixing the version of `click<8.0.0` next to where the kopf dependency is specified.

nolar:
[...]

PR #768, opened 2021-05-12
Adapt logging defaults to click 8.0

pshchelo:
also require at least this version of click

Fixes issue #767

## What do these changes do?

adapt logging to the new release of click

## Description

w/o this patch `kopf run` fails to start with click 8.0 without explicitly specifying `--log-format`

As kopf does not pin/cap click version, all users of already released versions will have to either cap `click<8.0.0` themselves or add explicit `--log-format full` to their `kopf run` arguments.

## Issues/PRs

 

> Issues: #767 

> Related:

## Type of changes

## Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,7 +45,7 @@ setup(
         'typing_extensions',    # 0.20 MB
         'python-json-logger',   # 0.05 MB
         'iso8601',              # 0.07 MB
-        'click',                # 0.60 MB
+        'click>=8.0.0',         # 0.60 MB
         'aiojobs',              # 0.07 MB
         'aiohttp<4.0.0',        # 7.80 MB
     ],
--- a/kopf/cli.py
+++ b/kopf/cli.py
@@ -40,7 +40,7 @@ def logging_options(fn: Callable[..., Any]) -> Callable[..., Any]:
     @click.option('-v', '--verbose', is_flag=True)
     @click.option('-d', '--debug', is_flag=True)
     @click.option('-q', '--quiet', is_flag=True)
-    @click.option('--log-format', type=LogFormatParamType(), default='full')
+    @click.option('--log-format', type=LogFormatParamType(), default=loggers.LogFormat.FULL)
     @click.option('--log-refkey', type=str)
     @click.option('--log-prefix/--no-log-prefix', default=None)
     @functools.wraps(fn)  # to preserve other opts/args


setup.py
```diff
@@ -45,7 +45,7 @@
         'typing_extensions',    # 0.20 MB
         'python-json-logger',   # 0.05 MB
         'iso8601',              # 0.07 MB
-        'click',                # 0.60 MB
+        'click>=8.0.0',         # 0.60 MB
```
nolar, line 48: Is the solution not compatible with 7.x.x?

pshchelo: have not tried it yet, lemme check

nolar:
Hm. It is interesting. I don't know what is happening with CI here. It looks like `pip` does not like the latest versions of some arbitrary packages unrelated to Click. E.g.:

```
Collecting pyngrok
  Downloading pyngrok-4.2.1.tar.gz (20 kB)
INFO: pip is looking at multiple versions of pyngrok to determine which version is compatible with other requirements. This could take a while.
  Downloading pyngrok-4.1.16.tar.gz (19 kB)
  Downloading pyngrok-4.1.15.tar.gz (19 kB)
  Downloading pyngrok-4.1.13.tar.gz (18 kB)
  Downloading pyngrok-4.1.12.tar.gz (18 kB)
  Downloading pyngrok-4.1.11.tar.gz (18 kB)
```

The broken jobs are reproducible (2 times):

* https://github.com/nolar/kopf/runs/2567772374?check_suite_focus=true

I have started a manual run of the same pipeline on the main branch — it works fine:

* https://github.com/nolar/kopf/actions/runs/836027521

It looks like an issue with the change, though I have no clue how Click could affect pyngrok, oscrypto, and other libraries.

pshchelo:
yes, very strange indeed. I am trying to reproduce that in the clean venv locally (w/o all those dev deps) and it stopped working as well... digging further

nolar:
Closing in favour of #769. Announced in https://github.com/nolar/kopf/issues/767#issuecomment-840495122

@pshchelo Thanks for your report and investigation on the issue! That made the fix much easier.

———

Regarding the CI — that might be caused exactly by the `click>=8.0.0` requirement. While fixing the issue in #769, I met this on my machine:

```
$ pip install click==8.0.0
………
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
import-linter 1.2.1 requires click<8,>=6, but you have click 8.0.0 which is incompatible.

$ pip --version
pip 21.1.1 from /Users/nolar/.pyenv/versions/3.9.1/envs/kopf391/lib/python3.9/site-packages/pip (python 3.9)
$ python --version
Python 3.9.1
```

The import-linter is used only for CI, not for runtime. And it has nothing to do with `certbuilder` or `pyngrok` which have failed in CI (they do not require Click). But it could be a trigger for `pip` to start reconsidering the whole solution of packages, which could cause the rescanning of these packages too (an unproven hypothesis).

PR #786, opened 2021-06-13
Upgrade to mypy 0.910 and fix the issues

nolar:
Upgrade mypy to 0.900. The main change is the extraction of stubs from mypy to separate libraries.

**On  mypy exclusion from pre-commit:**

Pre-commit is conceptually broken with mypy>=0.900. Mypy 0.900 has extracted some typed stubs from its own repo into separate libraries. Besides, Click 8.0.0 got its own type annotations instead of the mypy-embedded stubs.

Pre-commit uses its own isolated virtualenvs per hook. As a result, it cannot access the installed dependencies of the project's virtualenv. Which, in turn, breaks the mypy checks when executed in pre-commit's virtualenv. The decision not to fix this seems final: https://github.com/pre-commit/pre-commit/issues/730#issuecomment-835856309

This can be "solved" by adding `additional_dependencies: [...]` and listing all the dependencies which are already present in `requirements.txt` & `setup.py` (the hook also does not install the library itself; "-r" does not work due to wrong working dir). But this will be a code duplication and makes the dependencies harder to manage.

Instead, together with the upgrade, mypy is excluded from pre-commit and is executed on its own. Pre-commit remains only for small code-formatting things, e.g. whitespaces & co.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -41,14 +41,14 @@ jobs:
       - run: pip install -e .[${{ matrix.install-extras }}]
         if: ${{ matrix.install-extras }}
 
-      - run: mypy kopf --strict --pretty
+      - run: mypy kopf --strict
       - run: |
           # Mypying the examples
           exit_codes=0
           for d in $(find examples -maxdepth 1 -mindepth 1 -type d)
           do
             echo "Checking ${d}"
-            mypy $d --pretty
+            mypy $d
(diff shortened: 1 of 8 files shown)


nolar:
Builds with mypy 0.9xx are currently broken by import-linter which downgrades Click to `<8` (7.x.x has no type annotations). To be fixed in https://github.com/seddonym/import-linter/pull/98. This is also useful for better testing of Kopf vs. Click 8.x.x.

PR #792, opened 2021-06-23
Extract API sessions & HTTP requests as lower-level client routines

nolar:
We need to prepare the API client routines for the coming feature of retrying all API requests on server-side/networking errors. It is easier to implement such logic in one place rather than to apply it with decorators and/or composing objects in several places.

As such, the lower-level routines are extracted and all the API session management (authentication, response checking, etc) is shifted down the stack. The lower-level routines only do the generic API requests (GET/POST/PATCH/DELETE/etc) without knowing their purpose. The higher-level routines do additional object manipulation before sending anything to the lower-level API requests.

For convenience and code simplicity, most lower-level methods return a parsed object; all of them implicitly verify the response status.

_Extracted from #788 to reduce the diff size and remove unrelated changes._

### DIFF
--- a/kopf/_cogs/clients/api.py
+++ b/kopf/_cogs/clients/api.py
@@ -1,0 +1,221 @@
+import asyncio
+import json
+import ssl
+import urllib.parse
+from typing import Any, AsyncIterator, Mapping, Optional, Tuple
+
+import aiohttp
+
+from kopf._cogs.aiokits import aiotasks
+from kopf._cogs.clients import auth, errors
+
+
+@auth.authenticated
+async def get_default_namespace(
+        *,
+        context: Optional[auth.APIContext] = None,
+) -> Optional[str]:
+    if context is None:
+        raise RuntimeError("API instance is not injected by the decorator.")
+    return context.default_namespace
+
+
+@auth.authenticated
+async def read_sslcert(
+        *,
+        context: Optional[auth.APIContext] = None,
+) -> Tuple[str, bytes]:
+    if context is None:
+        raise RuntimeError("API instance is not injected by the decorator.")
+
+    parsed = urllib.parse.urlparse(context.server)
+    host = parsed.hostname or ''  # NB: it cannot be None/empty in our case.
+    port = parsed.port or 443
+    loop = asyncio.get_running_loop()
+    cert = await loop.run_in_executor(None, ssl.get_server_certificate, (host, port))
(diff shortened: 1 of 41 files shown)


kopf/_cogs/clients/api.py
```diff
@@ -0,0 +1,213 @@
...
+
+
+@auth.authenticated
+async def request(
+        method: str,
+        url: str,  # relative to the server/api root.
+        *,
+        payload: Optional[object] = None,
+        headers: Optional[Mapping[str, str]] = None,
+        timeout: Optional[aiohttp.ClientTimeout] = None,
```
psontag, line 46: Looking through the changes here I noticed that in most places there is no explicit timeout configured for the request being made. Is this purpose? Is the intention to rely on the [default timeout of 5min from aiohttp](https://docs.aiohttp.org/en/stable/client_quickstart.html#timeouts) in those cases? It might be nice to make this configurable for the user?

nolar: Good catch :-) The answer is: yes and no.

Yes — the timeouts are not configured and the default 5m timeout is used everywhere except the watch-streams.

No — the timeouts are to be added in the coming PR (together with or immediately after the PR with retrying on server-side/networking errors). I've just split the huge changeset (which exploded in complexity) into several PRs, one PR per one semantic change.

Something like this (it needs tests & docs, but it works in my draft):

```python
@dataclasses.dataclass
class NetworkingSettings:

    request_timeout: Optional[float] = 5 * 60  # == aiohttp.client.DEFAULT_TIMEOUT
    """
    A timeout for the entire duration of an API request (in seconds).

    The timeout is applied to all short requests (GET/POST/PATCH/DELETE/etc).
    It is not applied to continuous watch-streams (which are also GET);
    for the watch-streams, set one of
    ``settings.watching.client_timeout``/``settings.watching.server_timeout``.
    """

    connect_timeout: Optional[float] = None
    """
    A timeout for the connection & handshake of an API request (in seconds).
    """

    error_backoffs: Union[float, Iterable[float]] = (1, 1, 2, 3, 5, 8, 13, 21)
    """
    How many times and with which delays (seconds) to retry the API errors.

    For more information on the API error retrying, see :doc:`/resilience`.
    """
```

psontag: Awesome 👍

PR #793, opened 2021-06-30
Configure networking timeouts for API clients

nolar:
With this change, the timeouts can be configured for regular requests (cluster scanning, resource patching, etc). Specifically, two timeouts: for the whole request (default 5 minutes as before), and for the initial TCP connection (the default is not set, i.e. no timeout, but limited by the overall timeout of 5 minutes).

Previously, all requests except for the watch-streams were using the default timeout of aiohttp — which is 5 minutes — and it was not configurable.

Two exceptions:

* Watch-streams are configured with separate settings (`settings.watching`, not `settings.networking`).
* Auto-tunnelling webhook server always uses non-configurable default timeout (it was too difficult to make it configurable, and not worth the overcomplication for a utility dev tool).

Extracted from #788 for atomicity. Implicitly, this PR also passes the `settings` objects to all client routines.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -205,7 +205,29 @@ make sure to pre-scale the executor accordingly
 API timeouts
 ============
 
-Few timeouts can be controlled when communicating with Kubernetes API:
+Timeouts can be controlled when communicating with Kubernetes API:
+
+``settings.networking.request_timeout`` (seconds) is how long a regular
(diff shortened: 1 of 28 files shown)


nolar:
It was temporarily blocked by Kubernetes 1.16 & 1.17 issues. Unblocked in #794.

nolar:
Released as `kopf>=1.33.0`.

Issue #801, opened 2021-07-15
Looking for documentation on usage for kopf.Patch (with admissioncontroller)


cjbaar:
## Question

I am attempting to build an admission controller, using `on.mutate`. I am having trouble finding detailed documentation or examples of how to use the patch for something other than a top-level key. I have tried the following approaches.
```
def mutate1(name, patch: kopf.Patch, logger, **_):
    """test"""
    logger.info(f"Patching {name}")
    patch.spec['template']['metadata']['annotations']['test-value'] = '123'
```
```
def mutate1(name, patch: kopf.Patch, logger, **_):
    """test"""
    logger.info(f"Patching {name}")
    patch.spec['template']['metadata']['annotations'] = {'test-value': '123'}
```

Neither of these results in the annotation being added to my resource spec. What is the proper way to set a value through nested keys? Also, not sure why both of these approaches appear to fail, but the handler never logs anything other successes.
```
[INFO    ] 192.168.65.3 [15/Jul/2021:17:42:46 +0000] "POST /mutate1?timeout=10s HTTP/1.1" 200 311 "-" "kube-apiserver-admission"
```

## Checklist

- [x ] I have read the [documentation](https://kopf.readthedocs.io/en/latest/) and searched there for the problem
- [x ] I have searched in the [GitHub Issues](https://github.com/nolar/kopf/issues?utf8=%E2%9C%93&q=) for similar questions

## Keywords

patch, webhook

cjbaar:
OK, it looks like I had the handler mis-defined, and it was being filtered out. Got it running, and now I do see actual errors in the logs (webhook still returns 200 though, for some reason). The error is `KeyError: 'spec'`; I assume this comes from the use of `patch.spec`, which is what I see in the docs. So the question is still the same: is there more info on the use of the `patch` object? Thanks.

cjbaar:
FYI, I also tried this:
```
    patch = kopf.Patch({'spec': {'template': {'metadata': {'annotations': {'test-value': '123'} } } } })
```

This doesn't throw an errors in the logs, but also has no effect.

nolar:
The patch is a dict-like object. So, Python's tricks with `patch.setdefault('spec', {}).setdefault('template', {}).setdefault('metadata', {}).setdefault('annotations', {})['xyz'] = 'abc'` should work. All other Python tricks for dicts would work too.

You should use the `patch` object provided by the framework via kwargs. All newly created patch objects have no effect, as the framework is unaware of their existence.

[...]

PR #803, opened 2021-07-20
Modify JSONPatch to include `add` operation on missing keys

cjbaar:
## What do these changes do?

The original implementation of JSONpatch construction only allowed for `replace` operations. This did not work when attempting to patch a key that did not already exist. Comments in the code indicate this was intended to be implemented, but was never completed. This is my initial attempt at implementing the `add` operation.

## Description

* Attempt to complete the "to-do" list item previously located at https://github.com/nolar/kopf/blob/main/kopf/_cogs/structs/patches.py#L94
* Create new `original` property of the `Patch()` class to store original body used in comparison
* In admission controller request, copy `raw_body` to `patch.original`
* While iterating over patch keys, if not present, change to `add` method and use rest of dictionary as value
* Additional tests for multi-layer patches

This has been tested with patching appsv1/Deployment, adding layered annotations when the annotation key did not exist. Existing unit tests still succeed, though unsure of potential side effects to `delete` ops.

## Issues/PRs

https://github.com/nolar/kopf/issues/801

## Type of changes

## Checklist

### DIFF
--- a/kopf/_cogs/structs/patches.py
+++ b/kopf/_cogs/structs/patches.py
@@ -63,6 +63,7 @@ class Patch(Dict[str, Any]):
         self._meta = MetaPatch(self)
         self._spec = SpecPatch(self)
         self._status = StatusPatch(self)
+        self._original = {}
 
     @property
     def metadata(self) -> MetaPatch:
@@ -80,6 +81,14 @@ class Patch(Dict[str, Any]):
     def status(self) -> StatusPatch:
         return self._status
 
+    @property
+    def original(self):
+        return self._original
+
+    @original.setter
+    def original(self, value):
+        self._original = value
+
     def as_json_patch(self) -> JSONPatch:
         return [] if not self else self._as_json_patch(self, keys=[''])
 
@@ -92,5 +101,21 @@ class Patch(Dict[str, Any]):
                 result.extend(self._as_json_patch(val, keys + [key]))
         else:
             # TODO: need to distinguish 'add' vs 'replace' -- need to know the original value.
-            result.append(JSONPatchItem(op='replace', path='/'.join(keys), value=value))
+            print(self._is_in_original_path(keys))
+            if self._original and self._is_in_original_path(keys):
+                result.append(JSONPatchItem(op='replace', path='/'.join(keys), value=value))
+            else:
+                result.append(JSONPatchItem(op='add', path='/'.join(keys), value=value))
         return result
+
+    def _is_in_original_path(self, keys: List[str]) -> bool:
+        _search = self._original
+        for key in keys:
+            if key == '':
+                continue
+            try:
+                _search = _search[key]
+            except KeyError:
+                return False
+        return True
+        
--- a/kopf/_cogs/structs/patches.py
+++ b/kopf/_cogs/structs/patches.py
@@ -58,12 +58,12 @@ class StatusPatch(dicts.MutableMappingView[str, Any]):
 # Event-handling structures, used internally in the framework and handlers only.
 class Patch(Dict[str, Any]):
 
-    def __init__(self, __src: Optional[MutableMapping[str, Any]] = None) -> None:
+    def __init__(self, __src: Optional[MutableMapping[str, Any]] = None, body = {}) -> None:
         super().__init__(__src or {})
         self._meta = MetaPatch(self)
         self._spec = SpecPatch(self)
         self._status = StatusPatch(self)
-        self._original = {}
+        self._original = body
 
     @property
     def metadata(self) -> MetaPatch:
@@ -84,10 +84,6 @@ class Patch(Dict[str, Any]):
     @property
     def original(self):
         return self._original
-
-    @original.setter
-    def original(self, value):
-        self._original = value
 
     def as_json_patch(self) -> JSONPatch:
         return [] if not self else self._as_json_patch(self, keys=[''])
@@ -115,4 +111,3 @@ class Patch(Dict[str, Any]):
             except (KeyError, TypeError):
                 return False
         return True
-        
--- a/kopf/_cogs/structs/patches.py
+++ b/kopf/_cogs/structs/patches.py
@@ -81,10 +81,6 @@ class Patch(Dict[str, Any]):
     def status(self) -> StatusPatch:
         return self._status
 
-    @property
-    def original(self):
-        return self._original
-
     def as_json_patch(self) -> JSONPatch:
         return [] if not self else self._as_json_patch(self, keys=[''])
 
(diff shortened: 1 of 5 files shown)


kopf/_cogs/structs/patches.py
```diff
@@ -80,17 +81,38 @@ def spec(self) -> SpecPatch:
...
         return self._status
 
+    @property
+    def original(self):
+        return self._original
+
+    @original.setter
+    def original(self, value):
+        self._original = value
+
```
nolar: I am worried that an extra public signature is exposed here. Can we avoid this? For example, by instantiating the patch with an optional kwarg (positional args are taken by the dict-like signatures):

```python
patch = Patch()
patch = Patch(body=raw_body)
patch = Patch({...})
patch = Patch({...}, body=raw_body)
```

This will let the patch keeping the original body for reference, but not expose its existence to the users.

tests/admission/test_jsonpatch.py
```diff
@@ -21,13 +32,33 @@ def test_removal_of_the_key():
 
 def test_addition_of_the_subkey():
     patch = Patch()
+    patch.original = {'xyz': {'def': 456}}
+    patch['xyz'] = {'abc': 123}
+    jsonpatch = patch.as_json_patch()
+    assert jsonpatch == [
+        {'op': 'add', 'path': '/xyz/abc', 'value': 123},
```
nolar, line 39: Does it work, by the way? Or does it require `{op: add, path: /xyz, value: {}}` beforehand? I.e. the 2nd-level addition when the 1st level is absent — how is it interpreted by the webhook machinery from the Kubernetes side (I didn't try this case before)?

nolar (changes requested):
Super! Comparing against the original body will greatly improve the JSONPatch consistency.

I also thought of remembering all changes in a JSONPatch-like accumulator, but that seemed overcomplicated. This "body-way" is better indeed.

kopf/_cogs/structs/patches.py
cjbaar: Yeah; sounds good. I have updated the PR with changes to use kwargs here, and removed the `@setter` method.

tests/admission/test_jsonpatch.py
cjbaar: In my limited testing with actual k8s resources (in this case, deployments), the single add works. If we need more detailed test cases in the real world, might have to create a CRD or something that be tested with more levels. Also, I am currently running against 1.17 clusters, which is considered rather old at this point, so have not checked the behaviour on newer versions.

I have tested with the following. The output of the JSONPatch is shown above each result.

(a) The annotation key already exists:
```
{'op': 'replace', 'path': '/spec/template/metadata/annotations/xyz3', 'value': 'abc3'}
---
  template:
    metadata:
      annotations:
        xyz3: abc3
```

(b) The annotation key does not exist, but `annotations` does.
```
 {'op': 'add', 'path': '/spec/template/metadata/annotations/xyz3', 'value': 'abc3'}
---
  template:
    metadata:
      annotations:
        something: else
        xyz3: abc3
```

(c) The key `annotations` does not exist at all.
```
{'op': 'add', 'path': '/spec/template/metadata/annotations', 'value': {'xyz2': 'abc2'}}
---
  template:
    metadata:
      annotations:
        xyz2: abc2
```

nolar: Thanks for clarifying. That is what I wanted to know.

Speaking of annotations, I now wonder how it would handle annotations with "/" in them (e.g. `example.com/note: value`), and what is the supposed way of escaping slashes in JSONPatch. The internet says it is `~1` (and `~0` for the tilde itself). But this is a separate problem — let's not mix it here. It is better to do this as a separate PR.

kopf/_cogs/structs/patches.py
nolar: And what is the supposed used for the getter? (`body` kwarg is provided to handlers together with other body parts.)

My point was that deprecating and removing the public interfaces is much more difficult than just not adding them in the first place — if possible (i.e when there are better ways).

Besides the `.original`, LGTM.

nolar:
Ah! And the last thing — please add the sign-off to all your previous commits (you need to replay the history), for DCO. Please see: https://kopf.readthedocs.io/en/stable/contributing/#dco-sign-off — this can be done even manually by adding the line to the end of each commit message:

```
Signed-off-by: Your Name <your.email@example.com>
```

kopf/_cogs/structs/patches.py
cjbaar: Sorry, I guess am not completely sure what you are asking. Are you saying to get rid of the `original` property?

cjbaar:
Yeah I saw the notes in github on the signing, and followed the commands it suggested, but seems like it is not taking. Will try again.

kopf/_cogs/structs/patches.py
nolar: Yes, it would be better to remove it — unless there is a reason why it is exposed to users. (And to keep the protected `._original` field as it is.)

cjbaar: Got it; removed.

kopf/_cogs/structs/patches.py
```diff
@@ -58,11 +58,12 @@ def __init__(self, __src: "Patch") -> None:
 # Event-handling structures, used internally in the framework and handlers only.
 class Patch(Dict[str, Any]):
 
-    def __init__(self, __src: Optional[MutableMapping[str, Any]] = None) -> None:
+    def __init__(self, __src: Optional[MutableMapping[str, Any]] = None, body = {}) -> None:
```
nolar: Ugh. I missed this. Sorry. The strict typing requires a type here (see the [CI errors](https://github.com/nolar/kopf/pull/803/checks?check_run_id=3131928690)).

```python
    def __init__(
        self,
        __src: Optional[MutableMapping[str, Any]] = None,
        **,
        body: Optional[bodies.RawBody] = None,
    ) -> None:
```

Also above (in this alphabetical order, to satisfy isort/pre-commit):

```python
from kopf._cogs.structs import bodies, dicts
```

I guess, this will be enough. Please check with `mypy --strict kopf` (or via the CI).

nolar (changes requested):
All is good in general. Though, my bad — I overlooked a lack of type annotation. Here is the suggestion (and an import above should be fixed too).

kopf/_cogs/structs/patches.py
nolar: PS: Maybe `# type: ignore` is needed below on the line where you select a key from it (`_search[key]`). This would be fine, as it is explicitly caught there with TypeError.

cjbaar: Updated and verified with mypy

nolar:
Thank you for your contribution!

It is merged & pre-released as [`kopf==1.33rc2`](https://github.com/nolar/kopf/releases/tag/1.33rc2) — so that you can already try it. I'll make it into a release presumably on Sunday or Monday 25-26.07.2021.

PR #809, opened 2021-07-26
Update thirdparty.py for support kubernetes_asyncio.

### LINKED ISSUE #806: thirdparty support for kubernetes_asyncio?
### Problem

I am using the kubernetes_asyncio library to asynchronously call to Kubernetes APIs.

However, kopf.adopt(naturally harmonize_naming, adjust_namespace and label too) supports kubernetes.client.model but not kubernetes_asyncio.client.model.

https://github.com/nolar/kopf/blob/aee09c87d39dfb99a715783523bd20f4d30bc9b9/kopf/_cogs/helpers/thirdparty.py#L36

Since kubernetes_asyncio follows the model of the kubernetes python sdk as it is, there is no problem with compatibility.

### Proposal

I wish this project also supports kubernetes_asyncio.

### Code

```
if any(C.__module__.startswith('kubernetes.client.models.') for C in subcls.__mro__): 
    return True
```
It would be nice to change that part to
```
if any(C.__module__.startswith('kubernetes.client.models.') for C in subcls.__mro__): 
    return True
if any(C.__module__.startswith('kubernetes_asyncio.client.models.') for C in subcls.__mro__): 
    return True
```

### Additional information

_No response_

caffeinism:
The model of the official sdk of kubernetes and the model of kubernetes_asyncio work the same. So we can support the model in the same way.

closes: #806

### DIFF
--- a/kopf/_cogs/helpers/thirdparty.py
+++ b/kopf/_cogs/helpers/thirdparty.py
@@ -21,10 +21,16 @@ try:
 except ImportError:
     PykubeObject = _dummy
 
+V1ObjectMeta = V1OwnerReference = None
+
+try:
+    from kubernetes_asyncio.client import V1ObjectMeta as V1ObjectMeta, V1OwnerReference as V1OwnerReference
+except ImportError:
+    pass
 try:
     from kubernetes.client import V1ObjectMeta as V1ObjectMeta, V1OwnerReference as V1OwnerReference
 except ImportError:
-    V1ObjectMeta = V1OwnerReference = None
+    pass
 
 
 # Kubernetes client does not have any common base classes, its code is fully generated.
--- a/kopf/_cogs/helpers/thirdparty.py
+++ b/kopf/_cogs/helpers/thirdparty.py
@@ -35,6 +35,8 @@ class KubernetesModel(abc.ABC):
         if cls is KubernetesModel:
             if any(C.__module__.startswith('kubernetes.client.models.') for C in subcls.__mro__):
                 return True
+            if any(C.__module__.startswith('kubernetes_asyncio.client.models.') for C in subcls.__mro__):
+                return True
         return NotImplemented
 
     @property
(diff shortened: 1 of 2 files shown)


TahreemSaleem:
Hi, is there any timeline on when this change will be merged?

I had to use the V1Secret object from official k8s python client instead of the object provided by kubernetes_asyncio since the `adopt` function cannot accept the kubernetes_asyncio object(s)

TahreemSaleem:
Hi, is there any timeline on when this change will be merged?

I had to use the V1Secret object from official k8s python client instead of the object provided by kubernetes_asyncio since the `adopt` function cannot accept the kubernetes_asyncio object(s)

m3talstorm:
We also need this feature

tsyganov-ivan:
Sorry, but why this MR was closed? 
Supports of kubernetes_asyncio looks really important in async k8s operator.

nolar:
Let’s keep it open so that it remains in my attention. 

The nearest release is expected to happen with Python 3.12 support once the last blocker (iohttp) is released: #1067; plus all easy prs combined (and retested) in a single batch.

vmirz:
Also need this. @nolar Can this get pushed through?

kopf/_cogs/helpers/thirdparty.py
```diff
@@ -21,27 +21,16 @@ class _dummy: pass
 except ImportError:
     PykubeObject = _dummy
 
+V1ObjectMeta = V1OwnerReference = None
+
+try:
+    from kubernetes_asyncio.client import V1ObjectMeta as V1ObjectMeta, V1OwnerReference as V1OwnerReference
+except ImportError:
+    pass
```
nolar, line 29: These libraries provide independent subclasses for e.g. `V1ObjectMeta` or `V1OwnerReference`, which we create and populate internally. Even if the classes are drop-in replacements of each other, we should be very careful on what we pass to the libraries. 

As such, the imported classes are different, and the models below are different too.

nolar:
Closing in favour of a different PR with the full support for `kubernetes_asyncio`:

* #1283

PR #821, opened 2021-08-21
Handle connection errors during list_obj calls gracefully

### LINKED ISSUE #787: Watch stream dies without recovering
## Long story short
The Watch stream for a custom resource fails and does not get recreated again. This prevents the operator from seeing any events for that resource.

## Description

We have noticed that sometimes our Operator does no longer react to any resource changes on the cluster. The reason for this seems to be that the Watcher stream of the operator for that resource died without being reestablished. We were able to track down the following log message:

```bash
Watcher for <our-resource>@none has failed: Cannot connect to host 172.20.8.1:443 ssl:default [Connection reset by peer]
```
Restarting the operator after this log messages appears fixes the problem. Unfortunately we were no able to reproduce this behaviour yet and it only happens rarely. We will try to gather more information on our end and update this issue.

We are already using the following server and client timeouts:
```python
    settings.watching.server_timeout = 120
    settings.watching.client_timeout = 150
    settings.watching.connect_timeout = 5
```
So I don't think this is related to #204.

We assumed that kopf automatically restarts any failed watches automatically. Is this only the cases for certain kind of errors?
Is there any way to verify that all watchers are running? We could then use this logic as a workaround in our Operator health probe.

## Environment

* Kopf version: 1.32.1
* Kubernetes version: v1.19.11
* Python version: 3.8
* OS/platform: Debian

psontag:
This should catch any errors that occur during the initial listing of the objects, even those that make it through the retries. 
The listing should then be retried automatically like usual when the continuous watch ends.

I am not sure if this is the best way to fix this. I am not that familiar with the code base yet. So if you have any recommendations here let me know.

closes: #787

### DIFF
--- a/kopf/_cogs/clients/watching.py
+++ b/kopf/_cogs/clients/watching.py
@@ -153,10 +153,9 @@ async def continuous_watch(
         operator_pause_waiter: aiotasks.Future,
 ) -> AsyncIterator[Union[Bookmark, bodies.RawEvent]]:
 
-    where = f'in {namespace!r}' if namespace is not None else 'cluster-wide'
+    # First, list the resources regularly, and get the list's resource version.
+    # Simulate the events with type "None" event - used in detection of causes.
     try:
-        # First, list the resources regularly, and get the list's resource version.
-        # Simulate the events with type "None" event - used in detection of causes.
         objs, resource_version = await fetching.list_objs(
             logger=logger,
             settings=settings,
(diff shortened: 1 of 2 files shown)


nolar:
I've made several corrections regarding logging: to not log the reason of the stream restart — the same as it happens when the streaming API call fails, for consistency. The error is not silenced anyway — the API retrying routines will complain about the error as needed.

psontag:
Hey @nolar thanks for merging the PR 👍 .
Do you have an ETA for the next release?

nolar:
@philipp-sontag-by I can make an RC this weekend — I just want to run through all issues and see if anything else is easily doable — to release all the quick fixes at once (while trying to catch the joy of the last days of warm & sunny weather).

psontag:
Ok, no worries take your well earned time off 👍

nolar:
Pre-released as `kopf==1.34rc1` ([link](https://github.com/nolar/kopf/releases/tag/1.34rc1)) — can be installed via pip and checked.

Issue #729, opened 2021-04-01
Race Condition in Event Queuing When MODIFIED Arrive After CREATE but Before last-handled-configuration Was Written


paxbit:
## Long story short

MODIFIED events on just CREATED resources might arrive before last-handled-configuration was written. This leads to the MODIFIED event being treated as `Reason.CREATE` b/c its `old` version is still empty.

-------------------------
Loading the (empty) `old` manifest is tried here:
https://github.com/nolar/kopf/blob/1d657e2b1e1669466d68e783cc6d23819cd130ed/kopf/reactor/processing.py#L153

-------------------------  
Falsely setting the cause reason to `CREATE` as a result of the empty `old` manifest is done here:
https://github.com/nolar/kopf/blob/d29ac2bcb8481efd400910f36510f43dc1255851/kopf/reactor/causation.py#L197-L200

-------------------------
The handler is not being called b/c its cause does not match the resource changing cause:
https://github.com/nolar/kopf/blob/8fb507cc77ec55b179b7b115e40cef983f20b67b/kopf/reactor/registries.py#L201 

## Description

If the handler creating a resource via 3rd-party means like pykube still spends a small amount of time after creating the resource before returning, a quick update-after-create to the resource will queue up MODIFIED events before kopf had a chance to write its last-handled-config.

The following code snipped reproduces this. We had a situation where a 2-container pod had one container immediately crashing after creation. When this happened quickly enough after the pod was created the handler designated to deal with crashing containers was never called. Since I'm working from home via a DSL link to the data center where the cluster lives, the varying connection latency over the day through the VPN gateway is sometimes enough to trigger this. But only after today's lucky setting of a break-point (introducing a sufficient handler delay) right after the pod creation I was able to reliably reproduce it and find the root cause.

All the handler does btw after creating the pod is creating an event about the fact as well as setting kopf's `patch` dict. 

I believe this one to be broken at the queuing design level and have no good idea how to fix this. After looking at this I'm not sure the current implementation can be fixed for correctness without substantial rewrites (`memories`, maybe?). The assumptions currently made around last-handled-configuration can never be fully upheld as long as third parties other than kopf (a.k.a. pykube, kubernetes itself) modify resources too - which will of course always be true.
However I'd be very happily proven wrong. Maybe the already queued MODIFIED events sans kopf storage annotations can be augmented in-memory with the missing data by remembering the CREATED event long enough. IDK.

The following script:
1. Creates a pod with two containers. One of them crashes after 1s.
[...]

PR #829, opened 2021-09-10
feat: batch window suppression

paxbit:
Hi @nolar,

it's been a while since #729. Back then I monkey patched the topics discussed in the issue and went with that up till now. 

I'd like to split this in two parts of which this PR is the first. It disables the batch_window by default and allows to return to the previous behaviour on-demand. I believe the idea behind batch_window is viable in some cases but incorrect to assume for all operator by default. The doc string tries to explain the situation. 

It would be great if you could have a look at it.

Thanks!

### DIFF
--- a/kopf/_cogs/configs/configuration.py
+++ b/kopf/_cogs/configs/configuration.py
@@ -204,10 +204,22 @@ class BatchingSettings:
     How soon an idle worker is exited and garbage-collected if no events arrive.
     """
 
-    batch_window: float = 0.1
+    batch_window: float = None
     """
-    How fast/slow does a worker deplete the queue when an event is received.
-    All events arriving within this window will be ignored except the last one.
+    A lossy but in some cases viable optimization mechanism to take some 
+    load off of event processing.
+    Defines the debouncing interval a worker depletes the incoming event queue with.
+    All events arriving within this window will be ignored except the last one. This can have 
+    consequences depending on the type and use-case of the operator implemented using kopf. 
+    Example: Given an event sequence of resource A of ``CREATED > MODIFIED > DELETED``, if this 
+    sequence is received by the incoming event queue within ``batch_window`` any handlers dealing
+    with ``CREATED`` or ``MODIFIED`` events of resource A will never be called. Instead the first handler 
+    being called for A will be the ``DELETED`` handler, if one exists. Depending on what your operator 
+    does, this might be problematic. 
+    
+    If your use-case allows debouncing of event sequences for a resource, set this to ``> 0``. 
+    A notable side effect of doing this is an implicit handler delay for all handlers as all events 
+    will only be dispatched to the handler after this interval. 
     """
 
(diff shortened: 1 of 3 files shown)


nolar:
Thanks for this PR and for your time spent on it. However, I think, I will not merge it — but for a good cause:

While playing with two separate changes in separate branches, and after digging into the history of this "batching" solution, I have found that it was implemented for the only purpose — ensuring the consistency with the resource recently patched by Kopf itself. And this consistency was based on the assumption that the patched versions arrives into the queue fast enough. The assumption is wrong. It can take longer than 0.1 seconds to get the patched version.

Instead, resourceVersion should be used to ensure consistency (as got from the PATCH operation). I have this change already implemented, and now only struggle with auto-testing this case (asyncio tests are hard). 

Once the resourceVersion-based consistency is ensured, the "batching" is not needed at all: it is not a feature of Kopf, it was a hack in the first place. And so, I am going to remove it completely.

However, with no "batching" in place, the question remains how to process the quickly arriving changes of a resource (as mentioned in your PR's comments: `CREATED > MODIFIED > DELETED` as an example). I think the best way would be to handle them all, i.e. all the changes — even if it causes some extra CPU load. Batching was not a way of reducing the load — as implied in this quote — the load reduction was a side effect:

> A lossy but in some cases viable optimization mechanism to take some load off of event processing.

---

This is the 1st change of the mentioned two. The 2nd change is processing ALL incoming low-level events, including those considered "inconsistent" during those short time windows when the patch is performed but the patched version is not yet received. They must be excluded both from batching and from consistency logic. This requires some more sophisticated split of event- & change-processing — which I also have drafted, but not properly stabilised. It can take some time for me to finish those 2 PRs.

Meanwhile, regarding this PR and to ensure that you are not blocked: can you achieve this same effect by setting the batch window to `0`? As far as I can see into the internals of `asyncio.wait_for()`, it treats non-None timeouts which are `<=0` by timing out instantly. And so, the batching would not happen. It does not solve all the problems, but at least it should be equivalent to this PR's effects: the `if` condition is not needed in that case. Or do I miss something?

paxbit:
Hi @nolar,

Thanks for having a look! I think it is a good idea to get rid of batching altogether and I now understand a bit better what it was meant to solve in the first place. 

Regarding setting the batching window to `0`, I think I'll leave the monkey patches in place which did solve both issues fine for the last months. The patches are twofold.

1. Patch the `if batching_window is not None` into `worker`
2. Use the getters/setters of `progress` and `diffbase` to manage a simple cache for patches which have not made the full roundtrip so far. You can find a WIP of my attempt to translate the monkey patch into the fork here: https://github.com/paxbit/kopf/commits/feature/caching-state-storage/kopf/_cogs/configs

Both patches did solve our issues for the last months. You wrote above you have sth. in mind for the 2nd part. Can you make use of the caching idea or is it something you would rather disregard? 

Looking forward to the release integrating both topics :)

PR #834, opened 2021-09-20
Update `Mapping` type to match new `mypy`

sobolevn:
Hi!

I am working on a new feature in `mypy` and it identified a problem in your code.
`Mapping[str, object]` cannot be unpacked into `priority: int = 0, lifetime: int = 60, lastseen: Optional[str] = None,`, but `Mapping[str, Any]` can be.

Link: https://github.com/python/mypy/pull/11151

### DIFF
--- a/kopf/_core/engines/peering.py
+++ b/kopf/_core/engines/peering.py
@@ -116,7 +116,7 @@ async def process_peering_event(
         return
 
     # Find if we are still the highest priority operator.
-    pairs = cast(Mapping[str, Mapping[str, object]], body.get('status', {}))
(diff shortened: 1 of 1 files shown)


nolar:
Thanks for doing such checks pro-actively (it is an interesting idea!) and for sending this PR. 

Please ignore the errors with `DeprecationWarning: The loop argument is deprecated …` — I am fixing them right now in a separate PR (#828). I will merge this PR as soon as I fix those asyncio issues (maybe a rebase will be needed).

sobolevn:
Thanks a lot for the quick fix! 🤝

PR #843, opened 2021-10-01
Update hierarchies.rst

leroyshirto:
Fix small typos on the page


nolar:
It seems like the same was fixed with #910, which I have noticed a bit earlier while going through the list.

nolar:
It seems like the same was fixed with #910, which I have noticed a bit earlier while going through the list. Sorry.

PR #857, opened 2021-10-31
Add old, new, diff and operation to WebhookCause

### LINKED ISSUE #851: Allow using old and new for admission controllers
### Problem

There are often times when you want to know if the admission controller is interacting with an object "in update". It would be nice if an "in update" flag was exposed to the controllers along with the old body of the resource. It looks like kopf does have an "old" kwarg but for whatever reason that doesn't seem to work with an admission controller method i. e. on mutate or validate. 

### Proposal

Add the old kwarg to admission controllers 

### Code

_No response_

### Additional information

_No response_

sambhav:
Adding kwargs old, new, diff and operation to the WebhookCause and
passing them to the appropriate handler. This allows users to write
more powerful webhooks and allows them to create validation and
mutation webhooks that can look at the old object to determine
their behavior. This commit also adds the operation field to the
passed kwargs to allow users to create more powerful/flexibly webhooks.

closes: #851

### DIFF
--- a/kopf/_core/engines/admission.py
+++ b/kopf/_core/engines/admission.py
@@ -13,7 +13,7 @@ from typing_extensions import Literal, TypedDict
 from kopf._cogs.aiokits import aiovalues
 from kopf._cogs.clients import creating, errors, patching
 from kopf._cogs.configs import configuration
-from kopf._cogs.structs import bodies, ephemera, ids, patches, references, reviews
+from kopf._cogs.structs import bodies, diffs, ephemera, ids, patches, references, reviews
 from kopf._core.actions import execution, lifecycles, loggers, progression
 from kopf._core.intents import causes, filters, handlers, registries
 
@@ -126,6 +126,9 @@ async def serve_admission_request(
 
     memo = await memories.recall_memo(raw_body, memobase=memobase, ephemeral=operation=='CREATE')
     body = bodies.Body(raw_body)
+    old = bodies.Body(old_body) if old_body is not None else None
+    new = bodies.Body(new_body) if new_body is not None else None
+    diff = diffs.diff(old, new)
     patch = patches.Patch(body=raw_body)
     warnings: List[str] = []
     cause = causes.WebhookCause(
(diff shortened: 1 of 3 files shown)


kopf/_core/engines/admission.py
```diff
@@ -126,6 +126,9 @@ async def serve_admission_request(
 
     memo = await memories.recall_memo(raw_body, memobase=memobase, ephemeral=operation=='CREATE')
     body = bodies.Body(raw_body)
+    old = bodies.BodyEssence(old_body) if old_body is not None else None
+    new = bodies.BodyEssence(new_body) if new_body is not None else None
```
nolar: Thanks for the contribution. Generally, looks great and the feature might be useful. 

The only concern is here, in these lines. The body essence is different from the body — the essence removes all the garbage. The definition of "garbage" is flexible, but mostly it is the system metadata: e.g. generation/resourceVersion/timestamps. I'm not sure how Kubernetes passes the metadata in admission hooks — I simply didn't pay attention to it before. So, I will perform some experiments locally to see how it fits and feels before merging.

_As a side-note (no action is needed; for my own remembering later): In the change-detecting handlers, the essence also removes the status stanza and re-adds the explicitly monitored fields. But that was done for the purpose of not (over-)reacting to non-essential changes, i.e. to dummy/technical events or other controllers storing their status fields (or even annotations of other Kopf-based controllers). In the case of admission, the reaction happens anyway, so there is no big win in removing the status/annotations. It can actually be more confusing if a change is admitted, but the diff is empty — so, it should not be made empty. [/the end of the side-note]_

nolar:
PS: The massive unit-test failure is probably caused by 3rd-party changes in `async_timeout` — I will take a look at this later. It is not related to this PR.

```
DeprecationWarning: with timeout() is deprecated, use async with timeout() instead
```

nolar:
Please rebase on `main` and force-push — #859 has fixed the tests.

sambhav:
@nolar done :)

kopf/_core/engines/admission.py
sambhav: Would you prefer this to be the body or the body essence? I am good with either.

sambhav:
@nolar updated it to use bodies.Body instead. That should also fix the CI mypy issue.

sambhav:
@nolar just a ping in case this PR slipped through your inbox :)

nolar:
@samj1912 Sorry for the delay. Don't worry, I didn't forget it — just working hard on another issue (which finally works!).

nolar:
Released as [1.35.3](https://github.com/nolar/kopf/releases/tag/1.35.3)

sambhav:
Amazing! Thank you so much!

Issue #687, opened 2021-02-16
kopf.adopt How to set controller to False ? - multiple entries in ownerreferences section


sajuptpm:
Hi,
I am getting following error when i try to add multiple entries to ownerreference  section of custom resource using kopf.adopt function.
I am doing **kopf.adopt**  , then **patch_namespaced_custom_object** (update).
kopf.adopt , by default setting '**controller': True,**, i want to make it False

Error:
====
**Only one reference can have Controller set to true. Found \"true\" in references for bun-test2 and bun-test3","field":"metadata.ownerReferences"}]},"code":422}**

------------------

is there any way to add entries to ownerReferences with '**controller': False,** ??

{'ownerReferences': [{'**controller': False,**
          'blockOwnerDeletion': True,
          'apiVersion': 'kopf.dev/v1',
          'kind': 'KopfExample',
          'name': 'kopf-example-1',
          'uid': '6b931859-5d50-4b5c-956b-ea2fed0d1058'}]}},

https://github.com/nolar/kopf/blob/f20df84d4b1ee50aa7249cf6066826e5e2b31cf2/docs/hierarchies.rst

nolar:
Usually, the object is a plain dict. It can be modified. 

```python
kopf.adopt(...)
for ref in data['metadata']['ownerReferences']:
    ref['controller'] = False
```

`kopf.adopt()` is supposed to take full ownership of the object, and semantically, there can be only one owner.

Alternatively, you can add your own ownerReferences items with a fully controlled structure and content — without `kopf.adopt()`. Kopf does not do anything other than just injecting a little `dict` into that field. The current owner can be identified from the kwargs `name`, `namespace`, `uid`, and `resource` (owner refs need `resource.kind` and `f"{resource.group}/{resource.version}"` probably): https://kopf.readthedocs.io/en/stable/kwargs/#body-parts

sajuptpm:
Thanks nolar

PR #858, opened 2021-11-01
Allow setting controller block owner deletion false

nashant:
## What do these changes do?

This will allow setting controller and blockOwnerDeletion from both build_owner_references and append_owner_references

## Description

Previous behaviour:
Error from trying to set multiple owner controllers

New behaviour:
Allow explicitly setting `controller=False`, so no error

Does the change affect end users?
Only if they choose to use it

## Issues/PRs

> Related:
https://github.com/nolar/kopf/issues/687

## Type of changes

## Checklist

- [X] Unit tests for the changes exist
- [X] Documentation reflects the changes
- [X] If you provide code modification, please add yourself to `CONTRIBUTORS.txt`

### DIFF
--- a/kopf/_cogs/structs/bodies.py
+++ b/kopf/_cogs/structs/bodies.py
@@ -246,6 +246,7 @@ def build_object_reference(
 
 def build_owner_reference(
         body: Body,
+        *,
         controller: Optional[bool] = True,
         block_owner_deletion: Optional[bool] = True,
 ) -> OwnerReference:
--- a/kopf/_cogs/structs/bodies.py
+++ b/kopf/_cogs/structs/bodies.py
@@ -246,6 +246,8 @@ def build_object_reference(
 
 def build_owner_reference(
         body: Body,
+        controller: Optional[bool] = True,
+        block_owner_deletion: Optional[bool] = True,
 ) -> OwnerReference:
     """
     Construct an owner reference object for the parent-children relationships.
@@ -257,8 +259,8 @@ def build_owner_reference(
     for cluster resources, or e.g. ``apiVersion`` for ``kind: Node``, etc.
     """
     ref = dict(
-        controller=True,
-        blockOwnerDeletion=True,
+        controller=controller,
+        blockOwnerDeletion=block_owner_deletion,
         apiVersion=body.get('apiVersion'),
         kind=body.get('kind'),
         name=body.get('metadata', {}).get('name'),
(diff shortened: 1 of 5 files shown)


kopf/_cogs/structs/bodies.py
```diff
@@ -246,6 +246,8 @@ def build_object_reference(
 
 def build_owner_reference(
         body: Body,
+        controller: Optional[bool] = True,
+        block_owner_deletion: Optional[bool] = True,
```
nolar, line 251: Can you please make these arguments to be kwarg-only? As it is generally done for all "mode specifiers" in the public interface(s) — unlike for positional arguments which are typically the objects of the action, not its modes.

```suggestion
        *,
        controller: Optional[bool] = True,
        block_owner_deletion: Optional[bool] = True,
```

kopf/_kits/hierarchies.py
```diff
@@ -22,6 +22,8 @@ class _UNSET(enum.Enum):
 def append_owner_reference(
         objs: K8sObjects,
         owner: Optional[bodies.Body] = None,
+        controller: Optional[bool] = True,
+        block_owner_deletion: Optional[bool] = True,
```
nolar, line 27:
```suggestion
        *,
        controller: Optional[bool] = True,
        block_owner_deletion: Optional[bool] = True,
```

nolar (changes requested):
Would it also make sense to add these same-named arguments to `kopf.adopt()` — an aggregate on top of all these ownership/hierarchy functions? For consistency of the interfaces. Wdyt?

nolar:
Please rebase on `main` and force-push — #859 has fixed the tests.

nashant:
> Would it also make sense to add these same-named arguments to `kopf.adopt()` — an aggregate on top of all these ownership/hierarchy functions? For consistency of the interfaces. Wdyt?

I considered that and decided probably not. As per the comment you left on the issue I referenced I think adopt means that the child is under full control of the parent. If I'm adding an owner ref explicitly then I may well want to have multiple, if I'm adopting then I want that as the only owner

nashant:
I've rebased against main but am still getting the `tests/hierarchies/test_labelling.py` failures

nashant:
Ok, I've re-rebased against your main (forgot I had accidentally pushed to my main) and commits are now clean, but still the same test failures

nashant:
> > Would it also make sense to add these same-named arguments to `kopf.adopt()` — an aggregate on top of all these ownership/hierarchy functions? For consistency of the interfaces. Wdyt?
> 
> I considered that and decided probably not. As per the comment you left on the issue I referenced I think adopt means that the child is under full control of the parent. If I'm adding an owner ref explicitly then I may well want to have multiple, if I'm adopting then I want that as the only owner

@nolar what would you prefer?

PR #872, opened 2021-11-28
Upgrade to PyPy 3.8 & Kubernetes 1.22 in CI

nolar:
Depends on #874.


nolar:
Quickly reverted things are not considered merged :-) 

This is a mistake: the PyPy 3.8 tests are not passing yet. The main branch was reset to the previous commit and force-pushed in ≈1-2 mins.

PR #893, opened 2022-02-10
Suggested fix for issue #300

alexdyas:
"Print statement from example does not show up in logs for me but logging lib works"

Replaced `print` statement with `logging.info`, and modified output to reflect the change.

Use of `logging` is probably better practice than the print statement anyway.


nolar:
Can you please sign-off your commits? https://kopf.readthedocs.io/en/stable/contributing/#dco-sign-off

alexdyas:
Ok, signed I think.

PR #894, opened 2022-02-16
invocation: Warn when no more free workers

Frankkkkk:
Hi @nolar ,

Sync handlers and daemons are run inside a thread for each resource.
When the number of resources exceeds the default number of allocated
workers (for example `os.cpu_count() + 4` in python 3.9), then they are
kept pending and will thus never be launched.

As explained in the documentation, this number can be tweaked. On top of
that, this commit makes a warning appear if the limit is exceeded.

Signed-off-by: Frank Villaro-Dixon <frank.villaro@infomaniak.com>

### DIFF
--- a/kopf/_core/actions/invocation.py
+++ b/kopf/_core/actions/invocation.py
@@ -6,9 +6,11 @@ Also, decorated wrappers and lambdas are recognized.
 All of this goes via the same invocation logic and protocol.
 """
 import asyncio
+import concurrent.futures
 import contextlib
 import contextvars
 import functools
+import logging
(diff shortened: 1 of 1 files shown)


nolar:
Sorry for the delay in reviewing it.

I like the idea of making a warning when the pool limit is reached. There is a little nuance worth fixing or improving:

I have taken a look into ThreadPoolExecutor's source code: once the limit is reached, the number of threads never decreases, so the warning will be issued on every function call, even when the thread pool is idle/empty.

Can the warning be issued only once? Or only when the thread pool is really overloaded at the moment?

**PS:** Regarding the failed CI runs — please rebase on the recent main branch, it is all fixed there.

PR #917, opened 2022-04-28
Update url to metacontroller in the docs

Vad1mo:
The old link to https://metacontroller.app/ points to a shady website that is not about metacontroller. Updated with the current fork of metacontroller

Signed-off-by: Vadim Bauer <1492007+Vad1mo@users.noreply.github.com>

### DIFF
--- a/docs/alternatives.rst
+++ b/docs/alternatives.rst
@@ -5,52 +5,53 @@ Alternatives
 Metacontroller
 ==============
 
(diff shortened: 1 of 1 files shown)


Vad1mo:
why is the DCO bot still complaining?

nolar:
It seems like the same has been done by #928 a few moments ago. Thanks for pointing to the problem!

PR #919, opened 2022-04-28
Fix incorrect JSONPatch paths when special characters are used

### LINKED ISSUE #918: kopf doesn't construct JSONPatches correctly when keys include characters like `~` and `/`
This causes issues when trying to patch labels that contain `/` for eg `app.kubernetes.io/created-by`

See the offending lines at

https://github.com/nolar/kopf/blob/ab53ace82e62a6fa709bf5a580007eac1273ac34/kopf/_cogs/structs/patches.py#L94-L101

We need to ensure that `~` is encoded to `~0` and `/` to `~1`

See https://datatracker.ietf.org/doc/html/rfc6901#section-3 and https://github.com/json-patch/json-patch-tests/issues/42

sambhav:
Signed-off-by: Sambhav Kothari <sambhavs.email@gmail.com>

Fixes #918

### DIFF
--- a/kopf/_cogs/structs/patches.py
+++ b/kopf/_cogs/structs/patches.py
@@ -18,6 +18,14 @@ from kopf._cogs.structs import bodies, dicts
 JSONPatchOp = Literal["add", "replace", "remove"]
 
 
+def _escaped_path(keys: List[str]) -> str:
+    """Provides an appropriately escaped path for JSON Patches.
+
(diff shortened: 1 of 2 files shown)


sambhav:
@nolar could we please merge this and cut a release? This is a significant bug in the patching logic and is causing some issues. I am not sure why the pypy3.7 failure is happening. Seems unrelated to my fix.

PR #920, opened 2022-05-06
update diff data structure in docs

rubyisbeautiful:
Not sure what tinkering led to this situation, but got IndexError: tuple index out of range when the diff is e.g.: (('add', (), None, {'foo': 'bar'}),)


rubyisbeautiful:
- [x] update PR based on resolution of conversation in issue #921

Issue #926, opened 2022-06-15
Periodically refresh token used for Kubernetes api access


Christian-Schmid:
### Keywords

kopf, eks, service account token, kubernetes

### Problem

Hi, 

we're using kopf deployed to Kubernetes on AWS EKS service. 

With Kubernetes version 1.21 the service account token lifetime is limited to one hour (See "Important - BoundServiceAccountTokenVolume" section in [[1]](https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html#kubernetes-1.21)).
AWS EKS is extending this lifetime for the moment to 90 days, but is issuing a warning for pods still using such outdated tokens. 

While checking our pods, kopf was one of the services identified using stale tokens. 

I checked the ``kubernetes-client`` library used by kopf and they adapted their behaviour to support such short living tokens (by reading the mounted token file once every minute).

I've updated to the latest kopf / ``kubernetes-client`` version but the pod still shows up as one using outdated tokens. 

Is there a way to trigger / configure reload of the api token?

Looking into the code I found [[2]](https://github.com/nolar/kopf/blob/main/kopf/_core/intents/piggybacking.py#L91), which if I understand correctly is extracting the token only at startup and then continue to use it over the lifetime of the pod. 

Is this assumption correct?

Thanks 

Best regards

Chris
[1] https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html#kubernetes-1.21
[2] https://github.com/nolar/kopf/blob/main/kopf/_core/intents/piggybacking.py#L91

nolar:
Hello.

> … if I understand correctly is extracting the token only at startup and then continue to use it over the lifetime of the pod.

Not exactly. The behaviour of Kopf in regard to credentials is described here:

* https://kopf.readthedocs.io/en/stable/authentication/

Long story short, it starts the re-authentication cycle every time an API request returns HTTP 401. The API request that got 401 is postponed until new credentials are acquired (usually re-read from the filesystem) and then retried with those new credentials.

So, it is totally safe to expire the old tokens on the server (i.e. K8s API) side — Kopf will handle that. If Kopf could retrieve the tokens on the initial start, it will be able to retrieve them on re-auth.

---

However, once a token is retrieved, it will be used as long as possible until failed with 401. There is currently no functionality to renew the tokens proactively in advance. 

[...]

PR #933, opened 2022-07-11
Define credentials expiration time (optionally)

nolar:
The re-authentication happens on the same trigger as before: when no valid credentials are left in the vault when they are needed.

The same as before, there is no preliminary vault population or credentials refreshment, including on credentials expiration: although we know that the credentials have expired, the re-authentication does not happen until an API call is attempted.

The existing API calls initiated before the expiration —e.g. the watch-streams— will continue running and will not be interrupted until closed from the server side or timed out due to settings.

Addresses #926.

### DIFF
--- a/docs/authentication.rst
+++ b/docs/authentication.rst
@@ -33,6 +33,7 @@ To implement a custom authentication method, one or a few login-handlers
 can be added. The login handlers should either return nothing (``None``)
 or an instance of :class:`kopf.ConnectionInfo`::
 
+    import datetime
     import kopf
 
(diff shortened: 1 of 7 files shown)


asteven:
Not 100% sure yet, but I think this change causes all the credentials.ConnectionInfo instances created from functions in kopf/_core/intents/piggybacking.py to be initialized with an expiration date of datetime.datetime.max.

So they are never cleaned up and effectively DOS the operator as it is left re-trying with expired connectioninfo forever.

PR #953, opened 2022-08-31
Release memory when a deamon exits

jeff4747:
Release local cache of Kubernetes objects when a demon exits

If a daemon finishes, the live_fresh_body copy of the Kubernetes object won't be used again, as
long as no other daemons are referencing the same object.  But it live_fresh_copy is never set
to None, so it will not be garbage collected.

This patch will mark live_fresh_body for garbage collection when a deamon exits, as long as
no other running daemon is referencing the same object.

Test suite completed without errors (log attached)
[kopf-test-log.txt](https://github.com/nolar/kopf/files/9464774/kopf-test-log.txt)

### DIFF
--- a/kopf/_core/engines/daemons.py
+++ b/kopf/_core/engines/daemons.py
@@ -419,8 +419,21 @@ async def _runner(
         # Only the filter-mismatching or peering-pausing daemons can be re-spawned.
         if stopper.reason is None:
             memory.forever_stopped.add(handler.id)
-            # Save memory by forgetting the live body, since it will no longer be used.
-            memory.live_fresh_body = None
+
+        # If this daemon is never going to be called again, we can release the
+        # live_fresh_body to save some memory.
+        if handler.id in memory.forever_stopped:
+            # If any other running daemon is referencing this Kubernetes 
+            # resource, we can't free it
+            can_free = True
+            this_daemon = daemons[handler.id]
+            for running_daemon in memory.running_daemons.values():
+                if running_daemon is not this_daemon:
+                    if running_daemon.handler.selector.check(cause.resource):
+                        can_free = False
+                        break
+            if can_free:
+                memory.live_fresh_body = None
 
         # Save the memory by not remembering the exited daemons (they may be never re-spawned).
         del daemons[handler.id]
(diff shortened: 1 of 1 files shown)


nolar:
The original discussion: #951

kopf/_core/engines/daemons.py
```diff
@@ -420,6 +420,21 @@ async def _runner(
...
+        # If this daemon is never going to be called again, we can release the
+        # live_fresh_body to save some memory.
+        if handler.id in memory.forever_stopped:
+            # If any other running daemon is referencing this Kubernetes 
+            # resource, we can't free it
+            can_free = True
+            this_daemon = daemons[handler.id]
+            for running_daemon in memory.running_daemons.values():
+                if running_daemon is not this_daemon:
+                    if running_daemon.handler.selector.check(cause.resource):
```
nolar: Can you please remind me: what does this `if` condition do? Or, better phrased, what would be the case when the running daemons of this resource object X will not match the definition of the handler?

jeff4747: running_daemon is the other daemon.  If running_daemon and this_daemon are referencing the same Kubernetes object, then we can't free the body since it's still in-use by the other daemon.

Editing: I'm using this particular if just to confirm if they're pointing at the same Kubernetes object.  If there's a better check, I'm happy to change it.

nolar: Thanks for the clarification. I think this 2nd "if" is not needed. 

The cross-object isolation happens much earlier, a few architectural levels above, i.e. when the individual objects are split into separate processing streams:

* First, in `watcher()`/`worker()` using the `(resource, uid)` as the key (also see `get_uid()` there).
* Second, where the singular `memory` is extracted from the plural `memories` via `memories.recall()`.

When it comes to the `memory` & `memory.running_daemons`, it is already dedicated to a specific object. All other objects (of the same and other kinds) have their own `memory` struct.

So, to my understanding, the check for _"other daemons except for the current one"_ is sufficient (which is done on the line above). The selector check is redundant — it will always be `True` because of such a split.

So, this 2nd "if" should be removed. Otherwise, the PR is clear, I am ready to merge and release it ASAP.

jeff4747: if removed, new push coming shortly.

hapatrick:
What are the chances of this change getting into the next release of kopf? We've been running with a patched version of kopf with these changes for months now. We'd really like to be able to go back to using the standard release without having to patch.

hapatrick:
Thank you!

PR #965, opened 2022-10-18
fix(doc): Fix name of parameter

mboutet:
The name of the parameter in the doc for `kopf.on.validate` does not match the code. This PR updates the doc to match the code.

https://github.com/nolar/kopf/blob/fb10a44f5a5f6e5c458c4973c51395383303a8d6/kopf/on.py#L158


nolar:
Thanks for the fix!

PR #976, opened 2022-11-13
Replace dataclasses with attrs and slotted classes

nolar:
In preparation for further refactoring for proper time handling — so that this single change would be tested in all Python versions.

Besides, **attrs** produces the slotted classes by default, which is good for memory footprint, especially on the frequently used classes like `HandlerState`, `Outcome`, etc.

In all other aspects, it is essentially a drop-in replacement with some minor changes.

The `attrs` dependency is already installed for one of the sub-dependencies, so nothing new is essentially added to the deployment.

### DIFF
--- a/kopf/_cogs/configs/configuration.py
+++ b/kopf/_cogs/configs/configuration.py
@@ -26,15 +26,16 @@ the root object, while keeping the legacy names for backward compatibility.
     used interchangeably -- but so that it is understandable what is meant.
 """
 import concurrent.futures
-import dataclasses
 import logging
 from typing import Iterable, Optional, Union
+
+import attrs
 
 from kopf._cogs.configs import diffbase, progress
 from kopf._cogs.structs import reviews
 
 
-@dataclasses.dataclass
+@attrs.define
 class ProcessSettings:
     """
     Settings for Kopf's OS processes: e.g. when started via CLI as `kopf run`.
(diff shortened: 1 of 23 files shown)


nolar:
Closing for now as meaningless: the possible wins on the memory usage are dwarfed by the other memory footprint constituents:

* `kubernetes` library, if installed and imported, takes ≈55-60% of he operator memory, +20% for other modules, and only 20-25% going to the operator runtime; 1/3 of those 20-25% (so ≈10%) is the asyncio connection (TCP+SSL) structures. So, the pure operator structures taking only 10-15%.
* If `kubernetes` is not installed and not used, it does not make better: imports take 40%, 60% for runtime. Of those 60%, ≈1/3 are connections (TCP+SSL), so only ≈ 40% going to operator structures. 
  * Of those structures, the biggest one is `Stream` (for 1000 objects/workers), which is already a named tuple — cannot be optimized even more. The second biggest usage is the list of all the loaded objects in memory on the initial listing. Even if chunked, they all will be loaded, so no potential for extra gains. And then the plethora of smaller objects taking maybe 1/3 of all operator structures (so, ≈10-15% of the overall memory) — the only thing that can be optimized by `attrs`.

That memory usage is measured with `memray` in a simple operator with 1x on-create handler, executed on a cluster with 1000 pre-created objects (not handled during creation, all "new"), and filesystem-based storages (to avoid patching via the API). The handling of 1000 objects takes a few seconds, so not something to worry much, too.

PR #999, opened 2023-02-12
Fix updates code example

iTaybb:
Remove extra `kopf.adopt` line, and fix emphasizing.


nolar:
Note for self: adopting is documented in `deletion.rst`, so it was indeed too early to add this line in the `updates.rst`.

PR #1000, opened 2023-02-13
Fix crash when an APIResource has no verbs

lukasstockner:
When running a Kopf operator in a cluster that has a recent version of [KubeVirt](https://github.com/kubevirt/kubevirt) installed, it crashes on startup with `TypeError: 'NoneType' object is not iterable`.

The problem appears to be that KubeVirt registers an APIResource with no verbs, so the resource that is being fetched here contains `"verbs": None`, which causes the code in `_read_version` to fail.

There are actually quite a few such resources, but all the others contain slashes in their name so the filter in line 124 ignores them, which is why it worked before KubeVirt v0.58.1.

### DIFF
--- a/kopf/_cogs/clients/scanning.py
+++ b/kopf/_cogs/clients/scanning.py
@@ -118,7 +118,7 @@ async def _read_version(
                 ),
                 namespaced=resource['namespaced'],
                 preferred=preferred,
-                verbs=frozenset(resource.get('verbs', [])),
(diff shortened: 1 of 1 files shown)


nolar:
Sorry for the delay reviewing it. Released as 1.36.2. And thank you for the contribution!

lukasstockner:
> Sorry for the delay reviewing it. Released as 1.36.2. And thank you for the contribution!

No worries, thanks for the review!

PR #1015, opened 2023-03-20
Add Admission Request to admission handler parameters

### LINKED ISSUE #1002: Provide access to AdmissionRequest object in validate/mutate handler
### Problem

I'm writing an admission handler that needs access to the AdmissionRequest resource.resource field which does not appear to be available within the handler

### Proposal

Include the AdmissionRequest object as a parameter to the admission handlers so that all of the raw information is available to be accessed if needed

### Code

@kopf.on.validate('myresource', ignore_failures=True, operation='DELETE')
def myhandler(request, body, dryrun, spec, meta, patch, warnings, **_):
    print(f'handler is using {request['resource']['resource']} for something')

### Additional information

I'd be happy to work on a PR if that is helpful.

And thanks for this project - it's incredibly useful.

bobh66:
Adds the Admission Request to the function arguments for admission handlers.

closes: #1002

### DIFF
--- a/docs/kwargs.rst
+++ b/docs/kwargs.rst
@@ -323,6 +323,17 @@ unless they declare themselves as ``side_effects=True``.
 See more: :doc:`admission`.
 
 
+.. kwarg:: request
+
+Request
+-------
+
+`request` (``Map[str,Any]``) is the AdmissionRequest that triggered the admission
+handler.
+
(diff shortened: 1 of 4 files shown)


bobh66:
I was able to accomplish this using the dynamic kubernetes client.

PR #1054, opened 2023-09-14
Add WebHookDockerDesktopServer

sgaist:
Docker Desktop has been using the `host.docker.internal` host address for quite some and this can be used to help develop hooks locally in a fashion similar to k3d and Minikube.

This merge request adds the corresponding helper class so it makes developing kopf based operator/webhooks easier with it.


kopf/_kits/webhooks.py
```diff
@@ -582,6 +598,9 @@ async def guess_host() -> Optional[str]:
             return WebhookK3dServer.DEFAULT_HOST
         elif subject_cn == 'minikube' or issuer_cn == 'minikubeCA':
             return WebhookMinikubeServer.DEFAULT_HOST
+        elif any(alt_name for alt_name in certpath.first.subject_alt_name_value.native if "docker" in alt_name):
+            return WebhookDockerDesktopServer.DEFAULT_HOST
+
```
nolar: Overall, the PR is good, except for this place: it seems overly generic. Can the condition be narrowed to something more specific, which can be compared via `==` or `.endswith()`?

I tried installing Docker Desktop for Mac locally, and checking which SSL certs CNs it generates. It says:

```
$ openssl s_client -connect kubernetes.docker.internal:6443 </dev/null | openssl x509 -noout -ext subjectAltName
………
X509v3 Subject Alternative Name: 
    DNS:docker-for-desktop, DNS:kubernetes, DNS:kubernetes.default, DNS:kubernetes.default.svc, DNS:kubernetes.default.svc.cluster.local, DNS:kubernetes.docker.internal, DNS:localhost, DNS:vm.docker.internal, IP Address:10.96.0.1, IP Address:0.0.0.0, IP Address:127.0.0.1, IP Address:192.168.65.3
```

So, might guess is that `docker-for-desktop` is the only one deterministic for this product only (i.e. distinguishing it from other other kinds of "kubernetes on docker"). For a more generic case, `kubernetes.docker.internal` could be considered.

I will now correct this PR before merge. If you have cases when this might be not enough — please comment or report a new issue.

sgaist: I don't remember exactly why I used such a broad test but I think your change makes completely sense.

Thanks for the update !

Issue #785, opened 2021-06-11
Example deployment configs for webhooks


sambhav:
## Question

Currently there are is very little documentation on how to deploy an admission controller written in kopf.
It would be nice to update https://kopf.readthedocs.io/en/stable/deployment/ with this information.

Currently I am struggling to figure out which pieces I need to set for an admission controller to be properly deployed as a K8s native app.

## Checklist

- [x] I have read the [documentation](https://kopf.readthedocs.io/en/latest/) and searched there for the problem
- [x] I have searched in the [GitHub Issues](https://github.com/nolar/kopf/issues?utf8=%E2%9C%93&q=) for similar questions

## Keywords

webhook, admission controller, deployment

sambhav:
@nolar I would be happy to help add this to the docs, but I don't quite understand which parts kopf handles when deployed as a k8s native app. 

- Does it automatically generate the certs and update the webhook config?
- Does it need a service to be created?
- What config/settings should my webhook itself use? I wasn't sure which of the webhook servers were appropriate in this case.

nolar:
> I would be happy to help add this to the docs, …

I would be happy to review and comment on that! Though, I have a feeling that it will be a huge lengthy text worth a blog post or a series of posts or maybe a small book.

More on that, some "convenient" practices can be opinionated and considered less secure than needed. In this regard, I am not sure Kopf's documentation is a good place to give any recommendations on the security-related topics (this is out of the scope of the framework).

But at least I would be happy to add a link (or a few links) to such manuals or guides.

> … which parts kopf handles when deployed as a k8s native app.

Can you please clarify the question?

Essentially, you need to expose a port where Kopf listens from the pod, then create a service pointing to that pod, and then create (manually) `[Validating|Mutating]WebhookConfiguration` objects pointing to that service. I thought of Kopf generating them (codenamed "Kopf SDK"), but didn't prioritise this feature yet.

Besides, generate SSL certificates either way (e.g. via cert-manager, openssl, or some other CA — depending on the company policies). Put the CA to `caBundle`. Put the SSL cert+private key to the operator (I guess, as Secrets mounted as a volume).

Create an on-startup handler that sets the webhook server with all these certs & ports — as per the existing documentation.

Optionally, add some health checks (liveness/readiness probes).

In some environments and companies, cluster-local Ingress objects can be used for HTTPS endpoints instead of self-served HTTPS servers in the pod (in that case, the pod's server will be simple HTTP).

[...]

PR #1058, opened 2023-09-29
Feature multiple operations webhook

### LINKED ISSUE #1057: Lack of supporting multiple operations for one admission webhook endpoint
### Long story short

As we can see in the [official k8s docs](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/) we can have multiple operations for one webhook.
But in the kopf I realized we can not!
This is my workaround now and basically i am having 2 endpoints which is not necessary at all. I should have been able to wrap these two function in just one endpoint because as you can see the logic is the same.
```
@kopf.on.validate("kopfexample", operation="CREATE")
def create(spec, **_):
    try:
        create_or_update(spec)
    except kopf.AdmissionError as error:
        raise error

@kopf.on.validate("tempaccess", operation="UPDATE")
def update(spec, **_):
    try:
        create_or_update(spec)
    except kopf.AdmissionError as error:
        raise error


def create_or_update(spec):
    """ My core logic """
```
I've investigated more and found this line in the `kopf.on.validate`'s code.
```
operation: Optional[reviews.Operation] = None,  # -> .webhooks.*.rules.*.operations[0]
```
and this line on `admission.build_webhook` function
```
'operations': ['*'] if handler.operation is None else [handler.operation],
```
So is there a reason to do it like this? 

### Kopf version

1.36.2

### Kubernetes version

_No response_

### Python version

3.11

### Code

_No response_

### Logs

_No response_

### Additional information

_No response_

mehrdad-khojastefar:
Added support for multiple operations for one handler.
closes: #1057 
related: #785 

So now you can have handlers like this: 
`@kopf.on.validate("tempaccess", operation=["CREATE", "UPDATE"])`

Also it is backward compatible so you can use it like this too:
`@kopf.on.validate("tempaccess", operation="CREATE"` 

I have run tests using `pytest` and all of them passed.
I also add this commit to one of our projects and tested it manually. It successfully creates the validating webhook with multiple endpoints when managed option is set.

@nolar let me know it there is anything more that I must do. Thanks.


SmsS4:
This PR would help us in our web-hooks to remove duplication in our codebase.

nolar (approved):
Thanks for sending this PR! I have made a few adjustments:

`list[]` is a syntax from Python 3.9+, while Kopf has to support at least Python 3.8 (I will drop 3.7 in #1062 now). It should be `List[]`, not `list[]` from [PEP-585](https://peps.python.org/pep-0585/).

Besides, I suggest it to be any collection, be that a list, a tuple, a set, a frozen set, or something self-made. Hence, `Collection[]` as the most generic class with size & items but no ordering (which would make it a `Sequence[]`).

And also I prefer to have clear names about the content of that field, so it should be in plural, not singular. I have added `operations` (plural) everywhere and made `operation` (singular) supported but deprecated (with a warning — as for some other deprecated names, e.g. `namespace`/`namespaces`).

The `Operations` type is removed as too verbose — it can be expressed with StdLib's `typing` library + `Operation`, does not contain complexity like `Union[]`, so let's keep it simple.

mehrdad-khojastefar:
I can update the docs and examples as well on weekend.

PR #1100, opened 2024-01-30
fix the outdated doc for installing minikube on macos, brew cask inst…

Chanoian:
…all is not a valid brew command anymore

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -11,7 +11,6 @@ MacOS:
 brew install minikube
 brew install hyperkit
 
-minikube start --driver=hyperkit
 minikube config set driver hyperkit
 ```
 
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -8,12 +8,11 @@ use [minikube](https://github.com/kubernetes/minikube):
 MacOS:
 
 ```bash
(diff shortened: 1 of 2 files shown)


DEVELOPMENT.md
```diff
@@ -8,12 +8,11 @@ use [minikube](https://github.com/kubernetes/minikube):
...
 ```bash
-brew install docker-machine-driver-hyperkit
-sudo chown root:wheel /usr/local/opt/docker-machine-driver-hyperkit/bin/docker-machine-driver-hyperkit
-sudo chmod u+s /usr/local/opt/docker-machine-driver-hyperkit/bin/docker-machine-driver-hyperkit
+brew install minikube
+brew install hyperkit
 
-brew cask install minikube
-minikube config set vm-driver hyperkit
+minikube start --driver=hyperkit
```
nolar: Perhaps, this line should not be here — the cluster is started a few lines below in the next snippet.

nolar (changes requested):

DEVELOPMENT.md
Chanoian: yes you are correct. will fix this.

nolar (approved):

PR #1126, opened 2024-09-17
Introduce a FIPS compatible hashing algorithm with variable size digest

bravecorvus:
Replace blake2b with FIPS 140-2 compliant SHA2 hashing library when needed.

Signed-off-by: Andrew Lee <bravecorvus@users.noreply.github.com>

closes: #1125

### DIFF
--- a/kopf/_cogs/configs/conventions.py
+++ b/kopf/_cogs/configs/conventions.py
@@ -31,12 +31,52 @@ in most cases, they will be used as the annotation names with special symbols
 replaced; in some cases, they will be cut and hash-suffixed.
 """
 import base64
+import _hashlib
 import hashlib
 import warnings
 from typing import Any, Collection, Iterable, Optional, Set
 
 from kopf._cogs.structs import bodies, patches
 
+# Invoke hashlib algorithm based on the security mode.
+# In normal mode use the original blake2b based hashing.
+# If we in the restrictive FIPS mode, RHEL FIPS mode restricts
+# the hashlib functions like blake2 to use openssl blake2 implementations
+# which limit the parameters and hence doesn't allow to customise the digest size.
+# So we use shake_256 as alternative in FIPS mode which provides variable length
+# digest support and is an acceptable SHA-3 Algorithm.
+# This class writes a wrapper as the digest signature is different for both.
+# Credit to dushyantbehl who submitted this fix to aimhubio/aim in https://github.com/aimhubio/aim/pull/3217/commits/cf932f82a0042f1c2cd6bc6f1097a924b7e0606a which is the same problem
+class compliant_hash_algorithm:
+    digest_size: int = _HASH_SIZE
+    salt: int
+    is_fips_mode_enabled: bool
+    hashlib_state: None
+
+    # Based on the FIPS mode choose between blake2b or shake_256 hash function
+    def _invoke_hashlib(self):
+        if not self.is_fips_mode_enabled:
+            return hashlib.blake2b(digest_size=self.digest_size, salt=self.salt)
+        else:
+            return hashlib.shake_256()
+
+    def __init__(self, digest_size = None, salt = None):
+        if digest_size:
+            self.digest_size = digest_size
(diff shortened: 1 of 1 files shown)


kopf/_cogs/configs/conventions.py
```diff
@@ -31,12 +31,52 @@
 replaced; in some cases, they will be cut and hash-suffixed.
 """
 import base64
+import _hashlib
```
nolar, line 34: Using the internal libs would be a big red flag as it makes the library less cross-version or cross-python compatible (other implementations of Python can lack this library).

kopf/_cogs/configs/conventions.py
```diff
@@ -181,7 +221,7 @@ def make_v2_key(self, key: str, max_length: int = 63) -> str:
         return final_key
 
     def make_suffix(self, key: str) -> str:
-        digest = hashlib.blake2b(key.encode('utf-8'), digest_size=4).digest()
```
nolar, line 184: Frankly speaking, I do not even remember why it is blake2b here, since a usual md5 or sha256 would be more than fine. But it is as it is.

nolar (commented):
First of all, thank you for your contribution. After a thorough consideration and weighing all pros & cons, I nevertheless decided not to accept it for reasons stated below. But there is a workaround for what I think is the original problem, so I hope it does not impact your experience much, and provides a solution for those running on such restricted systems in the future.

The main argument against the change: changing the hashing algorithm will change the keys of all existing objects in the cluster, thus causing **massive re-processing of resources** on Kopf upgrade, including the duplicate "creation" event re-processing of objects that already exist there for a long time. This can be seen as highly undesired by the users of the framework, to say the least.

There was once an attempt to change the key format, and it led to a rather sophsticated logic for backwards compatibility, as seen here in this file as v1 vs. v2 keys. Any change in the hashing algorith would require introducing keys v3 and making a proper transition and support for both v1, v2, and v3 keys in the cluster, i.e. still supporting that old hashing logic.

I prefer avoiding extra complexity unless there is a significant benefit for a general "mass user".

---

As far as I understand, the problem is scoped to severely restricted FIPS-compliant systems. Sadly, I cannot estimate the share of such systems in the userbase of Kopf, even approximately. Until there is some data suggesting a high share, I assume it is low — based on the observation that the problem did not appear yet until recently, so the current state of things is "good enough".

I suggest that the users of these restricted systems instead use the extensibility provided by Kopf and inherit the storage classes `AnnotationsProgressStorage` & `AnnotationsDiffBaseStorage`, override the `make_v2_key()` or `make_suffix()` public methods, and set those in the operator's configuration.

Alternatively, they can switch to entirely different storage methods, such as sqlite, mysql, postgres, redis, or anything key-val of their choice, including the shared filesystem pseudo-database.

It is documented here:

* https://kopf.readthedocs.io/en/stable/configuration/#handling-progress
* https://kopf.readthedocs.io/en/stable/configuration/#change-detection

The storage switch is this simple:

```python

@kopf.on.startup()
def configure(settings: kopf.OperatorSettings, **_):
    settings.persistence.progress_storage = FipsAnnotationsProgressStorage()
    settings.persistence.diffbase_storage = FipsAnnotationsDiffBaseStorage()
```

In case a smooth transition is needed from the old to the new keys, additional instructions are available at:

* https://kopf.readthedocs.io/en/stable/configuration/#storage-transition

```python
import kopf

@kopf.on.startup()
def configure(settings: kopf.OperatorSettings, **_):
    settings.persistence.progress_storage = kopf.MultiProgressStorage([
        FipsAnnotationsProgressStorage(),
        kopf.AnnotationsProgressStorage(),
    ])
    settings.persistence.diffbase_storage = kopf.MultiDiffBaseStorage([
        FipsAnnotationsDiffBaseStorage(),
        kopf.AnnotationsDiffBaseStorage(),
    ])
```

The codebase of this PR can serve as a reference on how to implement these two storage classes with overridden methods for FIPS-compatible systems.

PR #1132, opened 2024-10-30
Add python 3.13 support

Meallia:
The  `logging.Handler` class changed in 3.13 and will now always try to acquire the lock in its `handle` method. (cf https://github.com/python/cpython/pull/109462)

Removed the optimization that created the handler without a lock.

Also configure the CI to run on Python 3.13.


nolar (approved):

nolar:
Thank you for your contribution! Sorry for a delayed response from my side. The change is released as `1.37.3`.

PR #1138, opened 2024-11-18
Fix the CI failures

nolar:


nolar:
#1132 fixes the same issue, but also upgrades Python, so closed in favour of that PR.

PR #1157, opened 2025-01-16
Apply networking timeout to aiohttp session globally

### LINKED ISSUE #377: [PR] Configure session timeous globally
> <a href="https://github.com/jumpojoy"></a>        A pull request by [jumpojoy](https://github.com/jumpojoy) at _2020-06-24 10:29:59+00:00_
> Original URL: https://github.com/zalando-incubator/kopf/pull/377
> &nbsp;

Previously timeouts were configured only for watching requests, this patch
make them configured directly in session constructor.

jumpojoy:
Use timeouts defined in networking settings for aiohttp session.

Closes: #377

### DIFF
--- a/kopf/_cogs/clients/auth.py
+++ b/kopf/_cogs/clients/auth.py
@@ -9,6 +9,7 @@ from typing import Any, Callable, Dict, Iterator, List, Mapping, Optional, TypeV
 import aiohttp
 
 from kopf._cogs.clients import errors
+from kopf._cogs.configs import configuration
 from kopf._cogs.helpers import versions
 from kopf._cogs.structs import credentials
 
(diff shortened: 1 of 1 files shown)


nolar:
Hello, @jumpojoy. Can you please describe what problem this PR solves?

Isn't this issue of timeouts for all API requests solved by these lines of code?

https://github.com/nolar/kopf/blob/dc2961b19be8c4f2584e70956838a0094dbbfe34/kopf/_cogs/clients/api.py#L61-L65

If not, what are the cases & scenarios where this pops up? I.e. what are those places where the API request is made outside of Kopf-to-Kubernetes internal layer, which already injects settings/timeouts on a per-request basis.

_I will close the PR for now, but will reopen when the issue is clear. Maybe it will require some remake._

jumpojoy:
@nolar thanks for reply I think this patch is more cosmetic sets timeout to whole session globally in case if we will have some request that not passing timeout directly.

PR #1176, opened 2025-07-01
chore(logging): update log level when retrying

msturdy-adaptive:
Use log level `WARNING` (instead of `ERROR`) when a request to the Kubernetes API fails **and will be retried**.  Reporting these cases as WARNING seems appropriate as the request _could_ still be successful.

### DIFF
--- a/kopf/_cogs/clients/api.py
+++ b/kopf/_cogs/clients/api.py
@@ -104,7 +104,7 @@ async def request(
                 logger.error(f"Request attempt {idx} failed; escalating: {what} -> {e!r}")
                 raise
             else:
-                logger.error(f"Request attempt {idx} failed; will retry: {what} -> {e!r}")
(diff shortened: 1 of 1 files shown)


nolar (approved):

nolar:
Usually errors are something that require attention and fixing the underlying cause. The K8s API requests should not fail normally. If they do, there is already a problem — even if the extra resilience layers make it non-critical this time (fatalities would be `logger.critical()`/`logger.fatal()`, which is +10 steps above `logger.error()`). So let's keep it as an error in the logs — bringing the developers' attention to the cause.

This also aligns with the log levels of errors coming from handlers in `execute_handler_once()`, where retryable temporary errors are nevertheless logged as `logger.error()` or `logger.exception()` — despite that the failed handlers can succeed eventually, but require human attention "just in case".

PR #1198, opened 2025-12-08
Enable LLM indexing of the docs

nolar:
Related:

* https://github.com/jacobtomlinson/sphinx-llm/issues/34

Currently, without the full rewrite of the docs, with no context or structure optimizations — that is a long task. Just pure technical files in this PR.


jacobtomlinson (commented):
This PR is exactly as straight forwards as I would hope it should be. Don't hesitate to open issues on `sphinx-llm` if any build problems come up in the future.

Issue #1087, opened 2024-01-09
Handle large resource spec being annotated to `last-handled-configuration`


Bharat23:
### Keywords

annotation, spec, last-handled-configuration, finalizers

### Problem
When dealing with a resource that has a large `spec`, the operator is throwing an error
```
APIClientError('Pod \"<REDACTED>\" is invalid: metadata.annotations: Too long: must have at most 262144 bytes'
```
This is due to the fact that it is trying to store the spec in the `last-handled-configuration` annotation. 
It is especially problematic during deletion as it blocks the deletion using the finalizer, which never gets removed. So, how does one go about configuring the operator to handle such a scenario?

Note:
As per the doc [Troubleshooting](https://kopf.readthedocs.io/en/stable/troubleshooting/)
Restarting the operator doesn't solve the problem. Only running this command unblocks it.
```bash
kubectl patch kopfexample kopf-example-1 -p '{"metadata": {"finalizers": []}}' --type merge
```

Ramblurr:
wow this is unfortunate, literally the first lines of code I write with kopf and I hit this error:

import kopf
import logging


```
@kopf.on.create("secret")
def create_fn(body, **kwargs):
    logging.info(f"A handler is called with body")
```


```
kopf run -n database kopftest.py
[2024-02-16 09:26:51,554] kopf._core.engines.a [INFO    ] Initial authentication has been initiated.
[2024-02-16 09:26:51,556] kopf.activities.auth [INFO    ] Activity 'login_with_kubeconfig' succeeded.
[2024-02-16 09:26:51,557] kopf._core.engines.a [INFO    ] Initial authentication has finished.
[2024-02-16 09:26:51,735] root                 [INFO    ] A handler is called with body
[2024-02-16 09:26:51,735] kopf.objects         [INFO    ] [database/sh.helm.release.v1.crunchy-postgres-operator.v1] Handler 'create_fn' succeeded.
[2024-02-16 09:26:51,736] kopf.objects         [INFO    ] [database/sh.helm.release.v1.crunchy-postgres-operator.v1] Creation is processed: 1 succeeded; 0 failed.
[...]

PR #1200, opened 2025-12-13
feat: Add support for removing fields from diff

mdnfiras:
related: https://github.com/nolar/kopf/issues/1087

This PR solves [this comment](https://github.com/nolar/kopf/issues/1087#issuecomment-2186077417) from the previously mentioned issue.

[trivy-dojo-report-operator](https://github.com/telekom-mms/trivy-dojo-report-operator)'s `VulnerabilityReport` resources can get very big and cause issues. Having the ability to remove a couple of fields from the diff solves it.

Acknowledgement: Thanks to @baracoder for suggesting the core idea behind this approach!

### DIFF
--- a/kopf/_cogs/configs/diffbase.py
+++ b/kopf/_cogs/configs/diffbase.py
@@ -91,7 +91,7 @@ class DiffBaseStorage(conventions.StorageKeyMarkingConvention,
                 try:
                     dicts.remove(essence, ignored_field)
                 except TypeError:
-                    # If the field is not present or does not support item deletion, just skip it.
+                    # If the field does not support item deletion, just skip it.
                     pass
 
         return cast(bodies.BodyEssence, essence)
--- a/kopf/_cogs/configs/diffbase.py
+++ b/kopf/_cogs/configs/diffbase.py
@@ -26,12 +26,15 @@ class DiffBaseStorage(conventions.StorageKeyMarkingConvention,
     https://kubernetes.io/docs/concepts/overview/object-management-kubectl/declarative-config/
     """
 
+    def __init__(self, ignored_fields: Iterable[dicts.FieldSpec] | None = None) -> None:
+        super().__init__()
+        self.ignored_fields = list(ignored_fields or [])  # materialize the iterable
+
     def build(
             self,
             *,
             body: bodies.Body,
             extra_fields: Iterable[dicts.FieldSpec] | None = None,
-            ignored_fields: Iterable[dicts.FieldSpec] | None = None,
     ) -> bodies.BodyEssence:
         """
         Extract only the relevant fields for the state comparisons.
@@ -86,13 +89,12 @@ class DiffBaseStorage(conventions.StorageKeyMarkingConvention,
         self.remove_empty_stanzas(cast(bodies.BodyEssence, essence))
 
         # Remove ignored fields if specified
-        if ignored_fields:
-            for ignored_field in ignored_fields:
-                try:
-                    dicts.remove(essence, ignored_field)
-                except TypeError:
-                    # If the field does not support item deletion, just skip it.
-                    pass
+        for ignored_field in self.ignored_fields:
+            try:
+                dicts.remove(essence, ignored_field)
+            except TypeError:
+                # If the field does not support item deletion, just skip it.
+                pass
 
         return cast(bodies.BodyEssence, essence)
 
@@ -125,9 +127,8 @@ class AnnotationsDiffBaseStorage(conventions.StorageKeyFormingConvention, DiffBa
             ignored_fields: Iterable[dicts.FieldSpec] | None = None,
             v1: bool = True,  # will be switched to False a few releases later
     ) -> None:
-        super().__init__(prefix=prefix, v1=v1)
+        super().__init__(prefix=prefix, v1=v1, ignored_fields=ignored_fields)
         self.key = key
-        self._ignored_fields = ignored_fields
 
     def build(
             self,
@@ -135,12 +136,7 @@ class AnnotationsDiffBaseStorage(conventions.StorageKeyFormingConvention, DiffBa
             body: bodies.Body,
             extra_fields: Iterable[dicts.FieldSpec] | None = None,
     ) -> bodies.BodyEssence:
-        essence = super().build(
-            body=body,
-            extra_fields=extra_fields,
-            ignored_fields=self._ignored_fields
-        )
-
+        essence = super().build(body=body, extra_fields=extra_fields)
         self.remove_annotations(essence, set(self.make_keys(self.key, body=body)))
         self.remove_empty_stanzas(essence)
         return essence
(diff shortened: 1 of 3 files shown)


kopf/_cogs/configs/diffbase.py
```diff
@@ -111,10 +111,12 @@ def __init__(
             *,
             prefix: str = 'kopf.zalando.org',
             key: str = 'last-handled-configuration',
+            remove_fields: Iterable[str] = None,
```
nolar: Can you please rename it to a single-word `ignored_fields`?

So that it is aligned with `extra_fields` in the same module & class, following the "adjective + noun" semantics instead of the imperative verb. (Any other appropriate adjective would work, too.)

kopf/_cogs/configs/diffbase.py
```diff
@@ -123,6 +125,18 @@ def build(
             extra_fields: Iterable[dicts.FieldSpec] | None = None,
     ) -> bodies.BodyEssence:
         essence = super().build(body=body, extra_fields=extra_fields)
+
+        # Remove extra fields if specified
+        for path in self._remove_fields:
+            d = essence
+            keys = path.split('.')
```
nolar: ~It is better to use `dicts.parse_field()` from `kopf._cogs.structs.dicts` — to keep the field specification conventions unified across the interfaces.~

Probably not relevant, since I recommended using `dicts.remove()` below, which accepts unparsed field specs.

kopf/_cogs/configs/diffbase.py
nolar: It is better to use `dicts.FieldSpec` from `kopf._cogs.structs.dicts` — to keep the field specification conventions unified across the interfaces. This also include the pre-parsed lists and tuples of individual field components.

`dicts` is already imported in this file.

kopf/_cogs/configs/diffbase.py
```diff
@@ -183,6 +199,17 @@ def build(
...
+            for k in keys[:-1]:
+                # If current parent field is not a dict then the path is invalid; skip removal
+                if not isinstance(d.get(k, {}), dict):
+                    break
+                d = d.get(k, {})
+            d.pop(keys[-1], None)
+
```
nolar: Should this functionality be moved to the parent class `DiffBaseStorage`, in order to avoid the code duplication?

kopf/_cogs/configs/diffbase.py
```diff
@@ -123,6 +125,18 @@ def build(
...
+        # Remove extra fields if specified
+        for path in self._remove_fields:
+            d = essence
+            keys = path.split('.')
+            for k in keys[:-1]:
+                # If current parent field is not a dict then the path is invalid; skip removal
+                if not isinstance(d.get(k, {}), dict):
+                    break
+                d = d.get(k, {})
+            d.pop(keys[-1], None)
```
nolar: I have a strong suspicion that this code semantically duplicates `dicts.remove()` from `kopf._cogs.structs.dicts`, which has an advantage of also removing the empty parents.

In that case, you can avoid writing numerous edge-case tests for the fields removal, as the removal is tested. Only a single test that the removal is happening/attempted at all would suffice.

tests/persistence/test_storing_of_diffbase.py
```diff
@@ -65,11 +65,21 @@ def test_annotations_storage_with_prefix_and_key():
     assert storage.prefix == 'my-operator.my-company.com'
     assert storage.key == 'diff-base'
 
+def test_annotations_storage_with_remove_fields():
+    storage = AnnotationsDiffBaseStorage(
+        remove_fields=['a.b', 'a.e.f.g', 'h', 'k.l.m']
+    )
+    assert storage._remove_fields == ['a.b', 'a.e.f.g', 'h', 'k.l.m']
```
nolar: It makes no sense on testing the private/protected fields. Depending on the intention, either make it public (as other fields), or do not test it.

However, a test that checks that the removal actually happens or is attempted is needed (as in: "test the promise, not the implementation").

nolar (changes requested):
Thanks for the contribution! The functionality might be helpful indeed. But it is better to use the existing routines to manage the dicts and fields uniformly with the rest of the framework.

kopf/_cogs/configs/diffbase.py
mdnfiras: `dicts.remove` produces an error `TypeError: 'str' object does not support item deletion` when the user tries to delete a field that is actually a string (or any type that doesn't support field deletion). For example trying to delete the field `a.b` will produce this error because `a` is a string and has no subfields:
```json
{
  "a": "b"
}
```
In my code I handle it with the check `if not isinstance(d.get(k, {}), dict):` and just skip it. also `d = d.get(k, {})` ensures that non existent fields are also skipped silently.
Do you think I should still catch these errors and handle them silently or let the errors propagate?

ps: thanks for the review!

tests/persistence/test_storing_of_diffbase.py
mdnfiras: Alright, I don't intend to access this field for now so I'm keeping it protected. I removed this kind of tests for this field.

kopf/_cogs/configs/diffbase.py
mdnfiras: I ended up silently ignoring non-existent fields, and fields that don't support item deletion.

mdnfiras: note: `dicts.remove` already ignores non-existent fields, so catching `TypeError` exception is just for ignoring fields that don't support item deletion.

tests/persistence/test_essences.py
```diff
@@ -178,3 +178,33 @@ def test_get_essence_clones_body(
...
+            },
+            'e': {
+                'f': 'g',
+            }
+        },
+        'h': {
+            'i': 'j',
+        }
+    })
+    storage = cls(ignored_fields=['a.b', 'a.e.f.g', 'h', 'k.l.m'])
```
nolar, line 199: Just for the record: `a.e.f.g` is not a field, only `a.e.f` is — so it is an extra test for the removal of a non-existent field, the same as `k.l.m`.

kopf/_cogs/configs/diffbase.py
```diff
@@ -26,6 +26,10 @@ class DiffBaseStorage(conventions.StorageKeyMarkingConvention,
     https://kubernetes.io/docs/concepts/overview/object-management-kubectl/declarative-config/
     """
 
+    def __init__(self, ignored_fields: Iterable[dicts.FieldSpec] | None = None) -> None:
+        super().__init__()
+        self.ignored_fields = list(ignored_fields or [])  # materialize the iterable
```
nolar, line 31: Note: I've added this materialization here, so that depletable iterables would persist and be reusable across multiple runs. I hope nobody uses generators (both expressions & functions) in this context, but just in case.

nolar (approved):

nolar:
Released as 1.39.1. Thanks for the contribution!

tests/persistence/test_essences.py
mdnfiras: Yes `a.e.f.g` is technically not a field, but this one in particular causes `dicts.remove` to throw a `TypeError` exception because `a.e.f` is a field but does not support item deletion, hence the extra test.

nolar: Got it. Makes sense, indeed.

PR #1202, opened 2025-12-19
refactor: just fix some typo - change function name from ochestrator to orchestrator (with r)

immanuwell:
I noticed some typo in 2 files - there was ochestrator instead of orchestrator (with r), so I fixed that
@nolar fyi


nolar (approved):

nolar:
Thanks for noticing! ;-)

PR #1216, opened 2026-01-23
Refresh webhook certificates at their expiration "notAfter" timestamp

### LINKED ISSUE #1135: Handling TLS cert refreshes
### Keywords

certs, tls, refresh

### Problem

Does the `kopf` webserver support automatic certificate refresh without requiring a server restart? For context, the `cert-manager` manages the certificates and mounts them into the pod running this webserver, but currently, the `kopf` server doesn't reload upon certificate refresh.

nolar:
An attempt to refresh the externally provided SSL certificates at their expiration time.

While the code is easy, there is one caveat: the usage of the hidden undocumented `ssl._ssl._test_decode_cert()` to get the SSL cert parsed into a dict — in order to retrieve its `notAfter`. There seems to be no way to do this in a proper way with StdLib — only with heavy-weight libraries like `cryptography` (which I previously tried to avoid at all costs for lightweight operators).

I'll keep this PR here — maybe I will find another way to parse the SSL certs later.

---

Closes #1135


kopf/_kits/webhooks.py
```diff
@@ -319,6 +330,13 @@ def _build_ssl(self) -> tuple[bytes | None, ssl.SSLContext | None]:
                 )
                 if cadata is None and self.certfile is not None:
                     cadata = pathlib.Path(self.certfile).read_bytes()
+
+                cert_dict = ssl._ssl._test_decode_cert(str(self.certfile))  # type: ignore[attr-defined]
```
nolar: A remark for self: using the very internal undocumented method is not good and unreliable.

nolar:
Seperseded by

* #1217

PR #1218, opened 2026-01-27
Fix the problem of hanging index handler at startup when worker_limit is less than number of resources

PhVHoang:
The changes introduced in [this commit](https://github.com/nolar/kopf/pull/661/changes/1d657e2b1e1669466d68e783cc6d23819cd130ed)
 caused the operator to hang indefinitely during startup when:

- An in-memory index is used, and
- worker_limit is set lower than the number of resources that need to be indexed.

In this scenario, the indexing process can never fully complete, which blocks the operator from progressing past initialization.

This PR fixes the issue by ensuring the indexing phase can complete even when the worker limit is lower than the total number of resources.

### DIFF
--- a/kopf/_core/reactor/processing.py
+++ b/kopf/_core/reactor/processing.py
@@ -98,14 +98,14 @@ async def process_resource_event(
                 logger=terse_logger,
             )
 
-            # Wait for all other individual resources and all other resource kinds' lists to finish.
-            # If this one has changed while waiting for the global readiness, let it be reprocessed.
+            # Mark this resource as indexed (drop its toggle from the wating set)
+            # This allows other workers to know this resource is done with indexing.
+            # IMPORTANT: we drop the toggle immediately after indexing, but don't wait here.
+            # The wait will happen inside process_resource_causes() before regular handlers run.
+            # This prevents deadlock when worker_limit < number of resources, as workers can
+            # return after indexing and pick up next resources from the queue.
             if operator_indexed is not None and resource_indexed is not None:
                 await operator_indexed.drop_toggle(resource_indexed)
-            if operator_indexed is not None:
-                await operator_indexed.wait_for(True)  # other resource kinds & objects.
-            if stream_pressure is not None and stream_pressure.is_set():
-                return
 
             # Do the magic -- do the job.
             delays, matched = await process_resource_causes(
@@ -120,6 +120,9 @@ async def process_resource_event(
                 memory=memory,
                 local_logger=local_logger,
                 event_logger=event_logger,
+                resource_indexed=resource_indexed,
+                operator_indexed=operator_indexed,
+                stream_pressure=stream_pressure,
             )
 
             # Whatever was done, apply the accumulated changes to the object, or sleep-n-touch for delays.
@@ -151,7 +154,20 @@ async def process_resource_causes(
         memory: inventory.ResourceMemory,
         local_logger: loggers.ObjectLogger,
         event_logger: loggers.ObjectLogger,
+        resource_indexed: aiotoggles.Toggle | None = None,
+        operator_indexed: aiotoggles.ToggleSet | None = None,
+        stream_pressure: asyncio.Event | None = None,
 ) -> tuple[Collection[float], bool]:
+    # Wait for all resources to be indexed before running regular handlers.
+    # IMPORTANT: skip this wait during tjhe initial indexing phase to avoid deadlock when worker_limit
+    # is set lower than the number of resources. We detect the initial indexng phase by checking 
+    # if resource_indexed is not None (it's only set during initial indexing in queueing.py:201).
+    # During initial indexing is complete, resource_indexed becomes None, and we wait to ensure
+    # handlers for new events have access to complete indices.
+    if operator_indexed is not None and resource_indexed is None:
+        await operator_indexed.wait_for(True)
+    if stream_pressure is not None and stream_pressure.is_set():
+        return ([], False)
 
     finalizer = settings.persistence.finalizer
     extra_fields = (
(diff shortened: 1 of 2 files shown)


tests/handling/indexing/test_blocking_until_indexed.py
```diff
@@ -92,9 +92,66 @@ async def delayed_readiness(delay: float):
...
+        memories=ResourceMemories(),
+        memobase=Memo(),
+        raw_event={'type': event_type, 'object': {}},
+        event_queue=asyncio.Queue(),
+        operator_indexed=operator_indexed,
+        resource_indexed=resource_indexed,
+    )
+    assert looptime == 0
+    assert set(operator_indexed) == {resource_listed}  # resource_indexed was dropped
+    assert handlers.event_mock.called
```
nolar, line 154: I am slightly lost here: these tests look equivalent: both their setup, their action, and their assertions are the same (or I do not see the difference). Can they be combined into one test?

kopf/_core/reactor/processing.py
```diff
@@ -98,14 +98,14 @@ async def process_resource_event(
...
+            # IMPORTANT: we drop the toggle immediately after indexing, but don't wait here.
+            # The wait will happen inside process_resource_causes() before regular handlers run.
+            # This prevents deadlock when worker_limit < number of resources, as workers can
+            # return after indexing and pick up next resources from the queue.
             if operator_indexed is not None and resource_indexed is not None:
                 await operator_indexed.drop_toggle(resource_indexed)
-            if operator_indexed is not None:
-                await operator_indexed.wait_for(True)  # other resource kinds & objects.
-            if stream_pressure is not None and stream_pressure.is_set():
-                return
```
nolar: I am a little bit confused here: what is the purpose of moving these two ifs to `process_resource_causes()`? That rotuine runs immediately after this code block here, there is nothing inbetween, so the move seemingly has no effect. What did I miss?

PS: Besides the extra `resource_indexed is None` check.

nolar (commented):
Thanks for the PR. I will try to reproduce it locally and see how the fix works.

~As I understand, the quickest workaround would be to declare all indexing handlers as `async def`, so that the threaded workers are not involved at all — is that correct?~

nolar:
> ~As I understand, the quickest workaround would be to declare all indexing handlers as `async def`, so that the threaded workers are not involved at all — is that correct?~

Please ignore this question. The `worker_limit` is not about threads, it is about the demultiplexer of watch-stream events. I've mixed this up with another setting.

nolar:
It was easy to reproduce by e.g. setting the `worker_limit=1` and creating 3x resources. Indeed, it blocks until the workers exit by `idle_timeout` (5 seconds).

However, applying the fix with the `operator_indexed` & `resource_indexed` changes does not solve the issue, it remains blocked with the same behaviour. — Can you please describe the principle behind the proposed solution?

---

I believe, if `worker_limit` is much lower than the number of resource objects in the cluster, the issue will remain with all types of handlers — because the core demultiplexer (the split of one watch-stream of the whole kind into queues of individual objects) will be capped by `worker_limit` + `idle_timeout` (and depending on the overall activity or inactivity of the cluster). There is no principal solution but to increase the limit. 

However, those other handler types do not block the operator startup as the indexing handlers do — maybe something can be done here, indeed.

PhVHoang:
Sorry @nolar for the late response. 

The principle of the fix is breaking a deadlock caused by a blocking wait during initial indexing.

The problem here (as I have reproduced in my local cluster) is during startup, each resource event worker calls `operator_indexed.wait_for(True)` inside `process_resource_event()`, which blocks until all resource kinds have
 finished indexing. When worker_limit < number of resources, all available workers get stuck waiting on this barrier, but the remaining resources in the queue can never be indexed because no workers are free to pick them up. 

The fix: Split the indexing into two phases by moving the blocking wait from `process_resource_event()` into `process_resource_causes()`:

  1. Initial indexing phase (`resource_indexed` is not None): The worker drops its toggle (signaling "I indexed this resource") and proceeds immediately to `process_resource_causes()` without waiting. This frees the worker to return to the queue and pick up the next resource.
  2. Post-indexing phase (`resource_indexed` is None): For events arriving after the initial listing is done, the wait is enforced inside `process_resource_causes()` before running regular handlers, ensuring they have access to complete indices.

nolar:
> 1. proceeds immediately to `process_resource_causes()` without waiting. This frees the worker to return to the queue and pick up the next resource.

@PhVHoang, thanks for clarifying. This quoted part is still unclear. As I see the code, `process_resource_causes()` immediately follows the triple-if block without any intermediate if-branches or interruptions. It means, from the point of view of the control flow, this move of the code changes nothing (it only matters from the point of functions as semantic abstractions).

That probably explains why I did not observe that the fix fixes the issue (when reproduced locally).

May I ask what was the original problem that you tried to solve with this fix? (As in the [XY Problem](https://en.wikipedia.org/wiki/XY_problem).) In other words, why do you want to have worker_limit set lower than the number of resources?

My biggest concern here is that this change will break the core promise: the handling should not start untill all indexes are fully built. This is what I wanted to check experimentally first.

nolar:
As a side note: I have merged the main branch into the feature branch, just in case. There were some core changes in the processing (1.42.0, #844).

nolar:
I am closing this PR for the following reasons:

- No response for some time;
- The problem remains unfixed with the proposed fix when reproduced locally (worker_limit=1);
- I do not understand the proposed solution and how it fixes what it declares as fixed;
- I have no other solution in mind without breaking the core promises in both indexing ("no handling until everything is indexed") and event delivery ("nothing is lost, dropped, or ignored").

In particular, exiting on the `stream_pressure` is unclear: even if we give way to the next event in the queue, it will be the same worker blocking the overall indexing, just on the next event.

I have added a precaution and workarounds to docs (keeping the reasonable `worker_limit` if indexing is used): 

* #1244 

Thank you for your effort anyway!

PR #1234, opened 2026-02-01
Parse proxy-url from kubeconfig or client libraries, pass to the session

### LINKED ISSUE #1076: kubeconfig proxy-url support
### Problem

I see a few issues around this topic... It looks like Kopf.ConnectionInfo doesn't support the proxy-url.
https://github.com/nolar/kopf/issues/187

### Proposal

_No response_

### Code

_No response_

### Additional information

_No response_

nolar:
The schema of kubeconfig's clusters:

* https://kubernetes.io/docs/reference/config-api/kubeconfig.v1/#Cluster

To test locally (MacOS):

```
brew install mitmproxy
```

Make `ca.pem` and `client.pem` from whatever certificates you use. This proxy cannot pass the client certificates through the connection. It accepts the client cert on the client<>mitm part, and presents its own client cert on the mitm<>server part. This is a limitation of the tool (probably).

```
mitmproxy --set ssl_verify_upstream_trusted_ca=ca.pem --set client_certs=client.pem
```

Add the mitm CA from `~/.mitmproxy/mitmproxy-ca-cert.pem` to the overall CA used by the client — because it fully substitues the SSL with its own CA on the client<>mitm part.

Add `proxy-url: http://127.0.0.1:8080` to the `~/.kube/config` for your current cluster, next to the `server: …` key.

TODOs:

- [x] Tests
- [x] Docs
- [x] Manual experiments (find a dev proxy server)

Closes #1076

### DIFF
--- a/kopf/_cogs/clients/auth.py
+++ b/kopf/_cogs/clients/auth.py
@@ -99,6 +99,7 @@ class APIContext:
                     ),
                     headers=info.as_http_headers(),
                     auth=info.as_aiohttp_basic_auth(),
+                    proxy=info.proxy_url,
                 )
             case credentials.AiohttpSession():
                 self.session = info.aiohttp_session
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,6 +4,7 @@ exclude: |
   (?x)^(
     docs/.*\.xml|
     docs/.*\.png|
+    tests/authentication/_proxy_server.py|
     tests/authentication/fixtures/pkey.pem
   )
 repos:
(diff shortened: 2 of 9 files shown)


kopf/_cogs/clients/auth.py
```diff
@@ -99,6 +99,7 @@ def __init__(
                     ),
                     headers=info.as_http_headers(),
                     auth=info.as_aiohttp_basic_auth(),
+                    proxy=info.proxy_url,
```
toelke, line 102: This parameter only exists with aiohttp >=3.11; but after #1233 aiohttp >=3.0.0 is required.

nolar: @toelke Do you have any issues upgrading aiohttp? I am going to require aiohttp>=3.11.0 by default (#1249), but can make it a conditional paremeters for <3.11.0 instead — if there are obstacles for upgrading.

That release is 1+ year old — somewhere in the border zone of freshness for deciding on which way to go.

toelke: No, we just updated aiohttp without a problem: We should be more on top of these things :-D

PR #1236, opened 2026-02-04
Fix finalizer race condition by switching to strategic merge patch

### LINKED ISSUE #1106: Finalisers race condition
### Long story short

As described [here](https://github.com/nolar/kopf/issues/798#issuecomment-1784635237).
I think that the description provided in that comment should be sufficient.

We are currently sometimes getting error -> `Forbidden: no new finalizers can be added if the object is being deleted`
when multiple controllers are handling the same resource.

### Kopf version

1.37.1

### Kubernetes version

1.27.3

### Python version

3.11.7

### Code

_No response_

### Logs

_No response_

### Additional information

_No response_

nolar:
## Summary

- Fixes #1106 — finalizer race condition when multiple controllers manage finalizers on the same resource
- Switches all K8s API patch calls from `application/merge-patch+json` to `application/strategic-merge-patch+json`
- `block_deletion()` now adds only the Kopf finalizer to the patch (strategic merge merges it into the existing list atomically)
- `allow_deletion()` now uses the `$deleteFromPrimitiveList/finalizers` directive to remove only the Kopf finalizer without touching others

## Background

With JSON merge patch, arrays are replaced wholesale. When Kopf removed its finalizer, it copied the full finalizers list, removed its own, and patched — overwriting any concurrent changes by other controllers. This caused `"Forbidden: no new finalizers can be added if the object is being deleted"` errors when multiple controllers operated on the same resource.

Strategic merge patch is safe for all existing behavior: metadata maps (labels, annotations) merge by key in both patch types, and CRD spec/status list fields default to atomic (replace) unless the CRD author explicitly opts into merge.

## Test plan

- [x] `tests/test_finalizers.py` — updated assertions for new patch structure (17 passed)
- [x] `tests/k8s/test_patching.py` — added content-type verification and directive cleanup test (22 passed)
- [x] `tests/handling/daemons/test_daemon_termination.py` — updated finalizer payload assertions (12 passed)
- [x] `tests/handling/test_multistep.py` — updated deletion-step assertion (24 passed)
- [x] `tests/reactor/test_patching_inconsistencies.py` — removed inapplicable finalizer inconsistency test case
- [x] Full broader test suite: 2072 passed, 0 regressions

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/kopf/_cogs/clients/patching.py
+++ b/kopf/_cogs/clients/patching.py
@@ -44,7 +44,7 @@ async def patch_obj(
         if body_patch:
             patched_body = await api.patch(
                 url=resource.get_url(namespace=namespace, name=name),
-                headers={'Content-Type': 'application/merge-patch+json'},
+                headers={'Content-Type': 'application/strategic-merge-patch+json'},
                 payload=body_patch,
                 settings=settings,
                 logger=logger,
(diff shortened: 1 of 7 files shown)


nolar:
Well, it took all the session tokens just to realize that there are no strategic merges in Kubernetes for most resources :-)

nolar:
Close as inappropriate, since there is no strategic merge in custom resource, only in the builtin ones, so it cannot be a generalised solution. Superseded by #1237. Kept here for memories.

PR #1237, opened 2026-02-04
Use JSON Patch for atomic finalizer operations

nolar:
## Summary

- Fixes the finalizer race condition (#1106) where `block_deletion()`/`allow_deletion()` copied the entire finalizers list and replaced it via JSON Merge Patch, overwriting concurrent changes by other controllers
- Finalizer add/remove operations are now accumulated in `Patch._finalizers_to_add`/`_finalizers_to_remove` and applied as a separate JSON Patch (RFC 6902) request after the merge-patch steps
- Uses optimistic concurrency via `resourceVersion` test ops; retries with fresh body on conflict (HTTP 422)

## Test plan

- [x] Verified on real k3d/k3s cluster: finalizer add on creation and remove on deletion both work correctly via JSON Patch
- [x] Test with multiple controllers modifying finalizers concurrently

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/kopf/_cogs/clients/errors.py
+++ b/kopf/_cogs/clients/errors.py
@@ -133,6 +133,10 @@ class APIConflictError(APIClientError):
     pass
 
 
+class APIUnprocessableEntityError(APIClientError):
+    pass
+
+
 class APITooManyRequestsError(APIClientError):
     pass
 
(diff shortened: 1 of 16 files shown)


nolar:
Works well, but overcomplicated. Preserved for history. Superseded by a simpler PR:

* #1252

PR #1241, opened 2026-02-07
Add static landing page for kopf.dev

nolar:


nolar:
Moved to https://github.com/kopf-dev/website

PR #1254, opened 2026-02-18
Use Protocol for function typing

sfc-gh-haryu:
Use [Python protocol](https://typing.python.org/en/latest/spec/protocol.html) rather than the [deprecated mypy extension](https://mypy.readthedocs.io/en/stable/additional_features.html#extended-callable-types).

This change makes it more compatible to users of other type checkers such as [ty](https://github.com/astral-sh/ty).

### DIFF
--- a/kopf/_core/intents/callbacks.py
+++ b/kopf/_core/intents/callbacks.py
@@ -9,8 +9,8 @@ a corresponding type or class ``kopf.Whatever`` with all the typing tricks
 (unions, optionals, partial ``Any`` values, etc) included.
 """
 import datetime
-from collections.abc import Callable, Collection
-from typing import TYPE_CHECKING, Any, TypeVar
+from collections.abc import Collection
+from typing import Any, Protocol, TypeVar
 
 from kopf._cogs.configs import configuration
 from kopf._cogs.helpers import typedefs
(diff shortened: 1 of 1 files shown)


nolar:
Related: #704 — where these types were converted from protocols to special mypy callable markers. Sadly, there are no remarks or explanation why it was done (the PR's goal is to make Kopf type-annotated properly). I vaguely remember some struggle with making those protocols work with example operators. It was in May 2021 (Python 3.9 as the latest released, >= 3.7 maintained) — so maybe protocols were not well-developed back then — but this is uncertain. I will double-check if it works now.

kopf/_core/intents/callbacks.py
```diff
@@ -9,264 +9,228 @@
...
-    ]
+# TODO:1: Split to specialised LoginFn, ProbeFn, StartupFn, etc. -- with different result types.
+# TODO:2: Try using ParamSpec to support index type checking in callbacks
+#         when PEP 612 is released (https://www.python.org/dev/peps/pep-0612/)
+class ActivityFn(Protocol):
+    def __call__(
+        self,
+        *,
+        settings: configuration.OperatorSettings,
+        index: ephemera.Index[Any, Any],
```
nolar: There is no kwarg `index`, the indexes come under their function names/ids (several of them are possible), so they fall under `**kwargs`. This line can be omitted, probably. There is no way to describe arbitrarily named kwargs in protocols (and this was a bad design anyway, in hindsight).

kopf/_core/intents/callbacks.py
```diff
@@ -9,264 +9,228 @@
...
-            DefaultNamedArg(Any, "param"),
-            KwArg(Any),
-        ],
-        invocation.SyncOrAsync[object | None]
-    ]
+class TimerFn(Protocol):
+    def __call__(
+        self,
+        *,
+        index: ephemera.Index[Any, Any],
```
nolar: Same here.

kopf/_core/intents/callbacks.py
```diff
@@ -9,264 +9,228 @@
...
+        name: str | None,
+        namespace: str | None,
+        patch: patches.Patch,
+        logger: typedefs.Logger,
+        memo: Any,
+        param: Any = ...,
+        **kwargs: Any,
+    ) -> invocation.SyncOrAsync[object | None]: ...
 
-    DelayFn = Callable[
```
nolar, line 200: `DelayFn` seemingly was lost in translation, it is used in a couple places, needs to be converted too.

nolar (changes requested):

kopf/_core/intents/callbacks.py
sfc-gh-haryu: Removed, thanks!

sfc-gh-haryu:
Thanks for the context. Wasn't aware that they were once Protocol.

nolar:
Sorry, I had to re-create the PR as #1281, since I do not have the push permissions to this origin. Your commits are preserved untouched, I have added one more to restore the lost type. Merged there.

Thank you for your contribution!

PR #1263, opened 2026-03-06
Add Python and Kubernetes health badges

Matheus-RR:
Adds Python and Kubernetes health badges from [ReleaseRun](https://releaserun.com) to the README. Since Kopf is a Python framework for Kubernetes operators, both badges are relevant. They show the current health grade of each platform, updated automatically based on version freshness, CVE status, and end-of-life timeline.

**Badge previews**:
Python Health Kubernetes Health

Free, no API key needed. Fits alongside the existing PyPI and CI badges.

Happy to adjust if you'd prefer different placement or only one of the two.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,8 @@
 [![Supported Python versions](https://img.shields.io/pypi/pyversions/kopf.svg)](https://pypi.org/project/kopf/)
 [![codecov](https://codecov.io/gh/nolar/kopf/branch/main/graph/badge.svg)](https://codecov.io/gh/nolar/kopf)
 [![coverage](https://coveralls.io/repos/github/nolar/kopf/badge.svg?branch=main)](https://coveralls.io/github/nolar/kopf?branch=main)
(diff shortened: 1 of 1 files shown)


nolar:
Hello. Thanks for the contribution. I do not get one thing about this service: how do these badges reflect the project-specific information? The links do not contain any references to Kopf. Or is it just the generic information about the latest Python per se and latest Kubernetes per se?

Matheus-RR:
Hey, good question. These are ecosystem-level badges, not Kopf-specific. They track the latest stable Python and Kubernetes versions (health grade based on CVE status, how fresh the release is, and EOL timeline). The idea is it gives contributors and users a quick signal about whether the core stack is in good shape. But yeah, they don't track anything specific to Kopf itself. If that doesn't feel like a good fit for the README, totally understand and happy to close this out.

Matheus-RR:
Going to close this out. We're reworking the badges to be repo-aware (detect actual versions from dependency files rather than showing generic ecosystem data). That would actually be useful in project READMEs, but this PR isn't that yet. Thanks for taking the time to look at it.

Issue #957, opened 2022-09-09
Handler stops responding after some time


mstrukci:
### Long story short

A handler of our custom resources stops responding after some time, despite having `connect_timeout` and `server_timeout` settings configured (referring to [#585](https://github.com/nolar/kopf/issues/585))

Is there any way to debug this issue? Can we setup a liveness probe for such handlers to get indicators when they are 'frozen'?

### Kopf version

1.35.4

### Kubernetes version

v1.22.8

### Python version

3.7.3

PidgeyBE:
Looks related to https://github.com/nolar/kopf/issues/955

We are seeing the same issue, on gcloud `v1.22.12` and kopf `1.35.4`.
The first time we saw it was after kopf was running for a few days.
The second time today was, by coincidence, at the same time a new node pool was added to our gke cluster.

PidgeyBE:
@mstrukci have you set `settings.watching.client_timeout` as well (as https://github.com/nolar/kopf/issues/585 suggests)?

StrukcinskasMatas:
Hey @PidgeyBE, we haven't set it initially. 
I've updated our operator with this setting and so far it has not frozen, however, the issue is not consistent, so it's hard to say if it isn't just a coincidence. I'll try to update this thread later after some more time passes.

PidgeyBE:
Hi @StrukcinskasMatas , my hypothesis is that it goes wrong when the kubernetes API is unavailable for some seconds (like in case a new node is added in Google cloud). Unfortunately I'm on (a long) holiday now and haven't had the chance to try to reproduce it.
If you would have any way to simulate some downtime in the kubernetes API, that would probably allow you to reproduce the issue...

StrukcinskasMatas:
Looks like even with setting the `client_timeout` attribute, the operator froze with the following event `event='Stopping the watch-stream for X cluster-wide.'`

We will probably look into working around this issue, by having a side service that sends events for the operator to process, and if it's not picked up, then restart the operator. In theory, this should solve our issue until it gets fixed by kopf.

bernardio:
Any updates on this issue?

PidgeyBE:
Seen this issue happening again now (in gcloud, I assume when scaling up a node), but because we have:
```python
    settings.watching.connect_timeout = 60
    settings.watching.server_timeout = 600
    settings.watching.client_timeout = 610
```
the operator started working again after 10 minutes...

james-mchugh:
[...]

Issue #847, opened 2021-10-12
controller built with kopf is stuck after being idle for 5-10 min  


shivamverma182:
### Long story short

I have created a controller which creates a custom object. It works fine as long as new custom objects are being created. But if it is idle for more than 5 min. the operator get hung and it does not respond to new custom objects being created.

I have checked its able to connect to kubernetes API server and also tried to set below.

@kopf.on.startup()
def configure(settings: kopf.OperatorSettings, **_):
    settings.watching.server_timeout = 300

### Kopf version

1.35.1

### Kubernetes version

1.20.9

### Python version

3.7

### Code

```python
import kopf
from kubernetes import client
import os
import yaml
import time


@kopf.on.startup()
def configure(settings: kopf.OperatorSettings, **_):
    settings.watching.server_timeout = 300


@kopf.on.create("subnamespacerequests")
def create_fn(spec, name, namespace, logger, **kwargs):

    labels = spec.get("labels")
    logger.debug(f"labels recieved: {labels}")
    annotations = spec.get("annotations")
    logger.debug(f"annotations receibed: {annotations}")

    path = os.path.join(os.path.dirname(__file__), "subnamespace.yaml")
    tmpl = open(path, "rt").read()
    text = tmpl.format(subnamespace=name, parentnamespace=namespace)
    data = yaml.safe_load(text)

    kopf.adopt(data)

    api = client.CustomObjectsApi()
    subns = api.create_namespaced_custom_object(
        group="hnc.x-k8s.io",
        version="v1alpha2",
        namespace=namespace,
        plural="subnamespaceanchors",
        body=data,
    )

    logger.debug(f"subnamespace {name} is created under namespace {namespace}")
    time.sleep(1)

    v1 = client.CoreV1Api()

    for item in v1.list_namespace().items:
        if item.metadata.name == name:
            logger.debug(f"Namespace {name} is created")
            ns = item

    logger.debug(f"existing labels: {ns.metadata.labels}")
    logger.debug(f"existing annotations: {ns.metadata.annotations}")
    ns.metadata.labels.update(labels)
    ns.metadata.annotations.update(annotations)
    patched_labels = ns.metadata.labels
    patched_annotations = ns.metadata.annotations
    logger.debug(f"patched labels: {patched_labels}")
    logger.debug(f"patched annotations: {patched_annotations}")
    patched_body = {
        "metadata": {"labels": patched_labels, "annotations": patched_annotations}
    }
[...]

Issue #698, opened 2021-02-24
Recovering from Missed Events


jeff-minard-ck:
## Question

We've created an operator with kopf and it's been great -- subscribe to our CRD creation/update/delete events and processing works most of the time, but only _most_. 

We routinely find ourselves having instances of our CRDS which have not been processed by the operator. If we go and make a change to the object (like adding an annotation), a new event is generated and kopf will then "see" the object and process it.

What it boils down to is that kopf is missing events -- but I don't think that's a kopf issue, in and of itself. Reading around suggests that events on watch streams are not inherently reliable. Several other operators I deal with handle this in with a two approach system:

1. `watch` for events and process them as needed
2. loop a `list` operation and reconcile items which have been missed

Even [the official docs on `watch`](https://kubernetes.io/docs/reference/using-api/api-concepts/#efficient-detection-of-changes) talk about needing to do option 2:

> Kubernetes [..] will only preserve a historical list of changes for a limited time. Clusters using etcd3 preserve changes in the last 5 minutes by default. When the requested watch operations fail because the historical version of that resource is not available, clients must handle the case by recognizing the status code 410 Gone, clearing their local cache, performing a list operation, and starting the watch from the resourceVersion returned by that new list operation.

`kopf` _does_ have something like this too; sorta. When the operator **starts up**, it does an initial load of all the objects and can setup all the handlers, etc.

Would it make sense to make that initial 'load' process something that could be invoked regularly?

## Checklist

- [x] I have read the [documentation](https://kopf.readthedocs.io/en/latest/) and searched there for the problem
- [x] I have searched in the [GitHub Issues](https://github.com/nolar/kopf/issues?utf8=%E2%9C%93&q=) for similar questions

## Keywords

missed events, missing events

nolar:
Thanks for asking. It is a good question. 

First, please clarify if those are indeed events missed or changes not detected — on.event vs. on.create/update/delete handlers. 

There is a known issue that changes made to an object while it is being processed are not always consistently handled: the earlier creation/update/deletion handlers might miss the most recent changes made within the same handling cycle (of multiple handlers) — #641 & co. 

---
If that is about on.event low-level handling, then it misses the events from the watch-streams intentionally if they arrive too fast. Only the latest state is processed. This is done because K8s, in general, is not an event-driven but a state-driven system. 

It can be configured by `settings.batching.batch_window` (default is 0.1 second). And, as I have just noticed, it is not fully documented (but should be). The source code can help for now:

[...]

Issue #614, opened 2020-12-14
Operator doesn't respond after some time


gilbeckers:
## Long story short
Right after the deployment of my controller, everything works fine: he reacts on create/delete/update events of resources. Then, after leaving the controller for some time (<10min), it stops handling changes of resources and nothing is logged anymore (also when running with `--verbose` ) 

## Description

While the controller seems to be frozen, my liveness probe continues to be working. Also, if I restart the controller, all changes that where made in the meantime, are handled. 
I tested it on kopf 0.27, 0.28 and 0.28.2, all version give the same kind of behavior. 

I must be doing something wrong, but I don't have any idea what can be wrong. I don't think that operator is still stuck in a handler, because the last lines of the logs says something like (which means that the handler is finished, right?)
```
[2020-12-14 16:34:02,413] kopf.objects [INFO    ] [sample-project] Handler 'update_members/spec.members' succeeded.
[2020-12-14 16:34:02,414] kopf.objects [INFO    ] [sample-project] Update event is processed: 1 succeeded; 0 failed.
```
Has someone seen this behavior before? Or any ideas on how I can debug this one? 

## Environment

* Kopf version: 0.27, 0.28 and 0.28.2
* Kubernetes version: 1.16
* Python version: 3.7
* OS/platform: Ubuntu

nolar:
@gilbeckers Can you please check #585 — does the solution from there works for you?

gilbeckers:
@nolar It seems that adding the client and server timeouts fixed my issue! Thanks for your help and many thanks for the amazing framework :+1:

nolar:
The issue should be fixed without any configuration in [1.44.0](https://github.com/nolar/kopf/releases/tag/1.44.0)

* #1267

PR #1267, opened 2026-03-09
Detect silent connection loss behind the cloud load balancers

nolar:
**Problem:** There is a well-known and frequent issue: the watch-stream sometimes becomes silent after 5-10 mins of inactivity, especially on the low-load clusters. This usually happens when any kind of cloud-hosted k8s is used like this: K8s<>LB<>Kopf.

The primary hypothesis (unproven) is that it is caused by the load balancer losing the connection K8s<>LB, but keeping the connection LB<>Kopf alive. As a result, Kopf (aiohttp, even the OS) does not know that the watch-stream is not alive anymore, and therefore does not reconnect.

**UPD:** _The users of Kopf reported a different analysis, where the connection is lost not because of a LB, which they do not have, but because of changing the internal IP addresses out of a pool of several of them. This does not change anything for Kopf: the effects are the same (silently dead connections), the fix is the same (inactivity tracking)._

* #957
* #585
* #614 
* #847 
* #698 
* #232 
* #204
* #860 (a failed attempt to fix it)
* maybe more, but I stopped searching

A helpful external issue:

* https://github.com/Azure/AKS/issues/1755
* https://github.com/nolar/kopf/issues/847#issuecomment-958963392

**Workaround:** A typical workaround is to set the client-side timeout to anything 1-10-20 minutes, but there is no good default that suits everyone — it depends on the cluster size and load. The downside is that we reconnect regardless of whether the connection is semi-dead or actually alive.

**Solution:** Now, it will activate k8s BOOKMARK events in the watch-stream, available since K8s 1.15 (the latest now is 1.35).

* https://kubernetes.io/docs/reference/using-api/api-concepts/#watch-bookmarks
* https://www.kubernetes.dev/resources/keps/956/

The bookmarks are hard-coded to happen every 60 seconds (currently non-configurable):

* https://github.com/kubernetes/kubernetes/blob/c20e0bc54189aef73a6a1498b4eab28b2457914f/staging/src/k8s.io/apiserver/pkg/storage/cacher/cacher.go#L62-L77

So, if there are no BOOKMARK events for longer than 60+ (e.g. 70) seconds, we can reliably conclude that the watch-stream is silently closed, so we can explicitly close it on our side and reconnect.

In the normal flow of operations, while the connection is alive, the BOOKMARK events will arrive, keeping the same original connection as long as possible.

**Scope:** Fixed only in Python 3.11+. It requires `asyncio.timeout()` and fails terribly with `asyncio.wait_for()` in tests. It is easier to wait ≈7 months till the Python 3.10 end-of-life (Oct 2026) than to build a complicated solution for it, which is already available in Python 3.11+.

Python 3.10 will work the old way — with potential freezes, ignoring the inactivity timeout. Nevertheless, Python 3.10 users can benefit from the BOOKMARK events flowing every 60 seconds and thus keeping the TCP connection alive — even without timeouts and reconnections.

**Difference:** This differs from the workaround with the same 60-second timeout (or any other default timeout) on the client connection, since we do not reconnect if the connection remains alive and the events are streaming. Frequent unnecessary reconnections were the main flaw in the workaround with timeouts.

**Side benefit 1:** Moreover, the mere existence of the BOOKMARK events might work as a keep-alive for the TCP connection (some traffic going), so it will not disconnect on the K8s<>LB part due to TCP inactivity — but this is a blind guess, no proof that it works or breaks that particular way, the safeguard for the reconnection is still needed.

**Side benefit 2:** Among other things, though not the goal, it helps keeping the latest resource version closer in time rather than the current resource version from the actual latest event — so the reconnections will use the time window of approximately 1 minute on the server side.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -273,6 +273,13 @@ The server-side timeouts are unpredictable, they can be 10 seconds or
 watching requests -- to prevent API flooding in case of errors or disconnects.
 The default is 0.1 seconds (nearly instant, but not flooding).
 
+``settings.watching.inactivity_timeout`` (seconds) is how long the watch stream
+is allowed to stay completely silent — delivering no events, not even bookmarks —
+before it is considered dead and closed for reconnection.
+Kubernetes sends bookmark events roughly every 60 seconds as a heartbeat,
+so the default of 90 seconds allows for reasonable jitter while still detecting
+streams that are silently stalled at the TCP level.
+
 .. code-block:: python
 
     import kopf
--- a/kopf/_cogs/clients/watching.py
+++ b/kopf/_cogs/clients/watching.py
@@ -21,6 +21,7 @@ import asyncio
 import contextlib
 import enum
 import logging
+import sys
 from collections.abc import AsyncIterator
 from typing import cast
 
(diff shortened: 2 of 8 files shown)


giannicarabelli:
Hi all. After some investigation, I think I've identified the root cause — applies to AWS EKS but likely elsewhere too.

## Context

In EKS there are no load balancers: AWS requires relying on FQDNs rather than IPs. Each zone has its own node, with its own IP, and those IPs are **not stable** — AWS can change them at any time.

## The problem

The `default/kubernetes` service holds the control plane IPs as endpoints (e.g. `x` and `y`). When AWS rotates them (e.g. `y` → `k`), the old node `y` can die **without cleanly closing TCP connections**, leaving them stale.

Under the hood, the service uses `iptables` for DNAT. When endpoints change, the iptables rules are updated — but **existing connections are unaffected** and keep believing they're alive.

Since the socket has **no `SO_KEEPALIVE`**, the connection hangs forever.

## Possible fixes

1. Enable `SO_KEEPALIVE` at OS level — but default kernel parameters are usually too large for this use case.
2. Enable `SO_KEEPALIVE` with **custom parameters directly in the code**, for finer control. I did this to monkey patch _aiohttp.TCPConnector._create_connection_  and seems ok.
3. Forget what's happening under the hood and use BOOKMARK to detect inactivity, as proposed.

PR #1110, opened 2024-03-18
consider task dependencies on shutdown

asteven:
The "credentials retriever" should not be stopped before other tasks, that depend on valid credentials, have been stopped.

closes: #1033

PR #1282, opened 2026-03-20
Terminate operator tasks sequentially in groups

### LINKED ISSUE #1033: peering depends on 'Credentials retriever' task but on shutdown that is stopped sooner
### Long story short

When using a ConnectionInfo with an expiration, which is required to work with expiring tokens, 
operator does not properly leave peering and does not exit.

It hangs as the updating/leaving the peering needs valid credentials but the vault ends up in a `need_reauth` state with no 'Credentials retriever' task left to populate the vault with new credentials.

### Kopf version

main

### Kubernetes version

any

### Python version

3.10.10

### Code

```python
@kopf.on.login()
async def authenticate(
        *,
        logger: kopf.Logger,
        **_: Any,
) -> Optional[kopf.ConnectionInfo]:

    try:
        kubernetes_asyncio.config.load_incluster_config()  # cluster env vars
        logger.debug("Async client is configured in cluster with service account.")
    except kubernetes_asyncio.config.ConfigException as e1:
        try:
            await kubernetes_asyncio.config.load_kube_config()  # developer's config files
            logger.debug("Async client is configured via kubeconfig file.")
        except kubernetes_asyncio.config.ConfigException as e2:
            raise kopf.LoginError("Cannot authenticate the async client library "
                                         "neither in-cluster, nor via kubeconfig.")

    # We do not even try to understand how it works and why. Just load it, and extract the results.
    # For kubernetes client >= 12.0.0 use the new 'get_default_copy' method
    if callable(getattr(kubernetes_asyncio.client.Configuration, 'get_default_copy', None)):
        config = kubernetes_asyncio.client.Configuration.get_default_copy()
    else:
        config = kubernetes_asyncio.client.Configuration()

    # For auth-providers, this method is monkey-patched with the auth-provider's one.
    # We need the actual auth-provider's token, so we call it instead of accessing api_key.
    # Other keys (token, tokenFile) also end up being retrieved via this method.
    header: Optional[str] = config.get_api_key_with_prefix('BearerToken')
    parts: Sequence[str] = header.split(' ', 1) if header else []
    scheme, token = ((None, None) if len(parts) == 0 else
                     (None, parts[0]) if len(parts) == 1 else
                     (parts[0], parts[1]))  # RFC-7235, Appendix C.

    #expiration = datetime.datetime.utcnow() + datetime.timedelta(minutes=1)
    expiration = datetime.datetime.utcnow() + datetime.timedelta(seconds=10)
    #expiration = None
    return kopf.ConnectionInfo(
        server=config.host,
        ca_path=config.ssl_ca_cert,  # can be a temporary file
        insecure=not config.verify_ssl,
        username=config.username or None,  # an empty string when not defined
        password=config.password or None,  # an empty string when not defined
        scheme=scheme,
        token=token,
        certificate_path=config.cert_file,  # can be a temporary file
        private_key_path=config.key_file,  # can be a temporary file
        priority=1,
        expiration=expiration
    )
```

### Logs

```none
^C[2023-06-27 21:58:49,358] kopf._core.reactor.r [INFO    ] Signal SIGINT is received. Operator is stopping.
[2023-06-27 21:58:49,358] kopf._core.reactor.r [DEBUG   ] Admission mutating configuration manager is cancelled.
[2023-06-27 21:58:49,359] kopf._core.reactor.r [DEBUG   ] Admission insights chain is cancelled.
[2023-06-27 21:58:49,359] kopf._core.reactor.r [DEBUG   ] Namespace observer is cancelled.
[2023-06-27 21:58:49,359] kopf._core.reactor.r [DEBUG   ] Credentials retriever is cancelled.
[2023-06-27 21:58:49,359] kopf._core.reactor.r [DEBUG   ] Admission webhook server is cancelled.
[2023-06-27 21:58:49,359] kopf._core.reactor.r [DEBUG   ] Admission validating configuration manager is cancelled.
[2023-06-27 21:58:49,360] kopf._core.reactor.r [DEBUG   ] Poster of events is cancelled.
[2023-06-27 21:58:49,361] kopf._cogs.clients.w [DEBUG   ] Stopping the watch-stream for customresourcedefinitions.v1.apiextensions.k8s.io cluster-wide.
[2023-06-27 21:58:49,361] kopf._cogs.clients.w [DEBUG   ] Stopping the watch-stream for clusterkopfpeerings.v1.kopf.dev cluster-wide.
[2023-06-27 21:58:49,363] kopf._cogs.clients.w [DEBUG   ] Stopping the watch-stream for netcenterips.v1alpha1.netcenter.hpc.ethz.ch cluster-wide.
[2023-06-27 21:58:49,363] kopf._cogs.clients.w [DEBUG   ] Stopping the watch-stream for services.v1 cluster-wide.
[2023-06-27 21:58:49,363] kopf._cogs.clients.w [DEBUG   ] Stopping the watch-stream for ingresses.v1.networking.k8s.io cluster-wide.
[2023-06-27 21:58:49,363] kopf._core.reactor.r [DEBUG   ] Daemon killer is cancelled.
[2023-06-27 21:58:49,363] kopf._core.reactor.r [DEBUG   ] Resource observer is cancelled.
[2023-06-27 21:58:59,370] kopf._core.reactor.o [DEBUG   ] Streaming tasks are not stopped: finishing normally; tasks left: {<Task pending name='peering keep-alive for default@None' coro=<guard() running at ./kopf/kopf/_cogs/aiokits/aiotasks.py:108> wait_for=<Future pending cb=[shield.<locals>._outer_done_callback() at /usr/lib/python3.10/asyncio/tasks.py:864, Task.task_wakeup()]>>}
[2023-06-27 21:59:09,379] kopf._core.reactor.o [DEBUG   ] Streaming tasks are not stopped: finishing normally; tasks left: {<Task pending name='peering keep-alive for default@None' coro=<guard() running at ./kopf/kopf/_cogs/aiokits/aiotasks.py:108> wait_for=<Future pending cb=[shield.<locals>._outer_done_callback() at /usr/lib/python3.10/asyncio/tasks.py:864, Task.task_wakeup()]>>}
[2023-06-27 21:59:19,386] kopf._core.reactor.o [DEBUG   ] Streaming tasks are not stopped: finishing normally; tasks left: {<Task pending name='peering keep-alive for default@None' coro=<guard() running at ./kopf/kopf/_cogs/aiokits/aiotasks.py:108> wait_for=<Future pending cb=[shield.<locals>._outer_done_callback() at /usr/lib/python3.10/asyncio/tasks.py:864, Task.task_wakeup()]>>}
... for a long time until finally
./run: line 8: 2849365 Killed                  kopf run --all-namespaces $@ ./handlers.py
```

### Additional information

This only happens with peering enabled and when using a ConnectionInfo with expiration set.

nolar:
**Problem:** We now cancel all tasks at once (as soon as any task exits, e.g. an exit-flag waiter). However, some tasks require other tasks to properly terminate. For instance, the peering shutdown in the orchestrator requires the authenticator to continue working. With the current approach, the authenticator exits soon, and the peering cannot shut down and de-register itself properly.

**Solution:** With the new approach, tasks are nested in groups, each group being a task itself. The termination usually starts from the most nested group, and unfolds inside out, terminating the inner groups, then middle groups, then outer groups.

**Limitation:** By this, we break the existing interface where all tasks are returned flat. We replace it with a list of one and only one task — the outermost group. This keeps the public interface intact, assuming the users did what we expected them to do: `spawn_tasks()`, and then pass that list down to `run_tasks()` unmodified. Alternatively, the users could orchestrate tasks their own way — this will work, too (with a list of a single root task). The exposure to the tasks structure will disappear though.

The legacy flat list can be requested (I hope no one needs it, but it is there just in case): `spawn_tasks(…, flat=True)`.

Fixes #1033. Supersedes #1110. Replaced by #1286.

TODOs:

- [ ] Revise the termination in edge cases: regular errors, hung tasks, etc. I have lost it for a moment.
- [ ] Review the timeouts (cancellation, exit) and intervals — to be the same as before the change.
- [ ] Give task groups better names — for log readability (I will have to debug it from logs one day).
- [ ] Maybe: adjust the docs to reflect the proper termination sequence (do we doc it at all?).
- [ ] Adjust tests and add new ones.
- [ ] Remove debug logs (🔥).

### DIFF
--- a/kopf/_core/reactor/running.py
+++ b/kopf/_core/reactor/running.py
@@ -5,6 +5,7 @@ import signal
 import threading
 import warnings
 from collections.abc import Collection, Coroutine, MutableSequence, Sequence
+from typing import Any
 
 from kopf._cogs.aiokits import aioadapters, aiobindings, aiotasks, aiotoggles, aiovalues
 from kopf._cogs.clients import auth
@@ -158,6 +159,7 @@ async def spawn_tasks(
         ready_flag: aioadapters.Flag | None = None,
         vault: credentials.Vault | None = None,
         memo: object | None = None,
+        flat: bool = False,  # for backwards compatibility, just in case
         _command: Coroutine[None, None, None] | None = None,
 ) -> Collection[aiotasks.Task]:
     """
@@ -196,7 +198,9 @@ async def spawn_tasks(
     signal_flag: aiotasks.Future = asyncio.Future()
     started_flag: asyncio.Event = asyncio.Event()
     operator_paused = aiotoggles.ToggleSet(any)
-    tasks: MutableSequence[aiotasks.Task] = []
+    inner_tasks: list[aiotasks.Task] = []
+    authn_tasks: list[aiotasks.Task] = []
+    outer_tasks: list[aiotasks.Task] = []
 
     # Map kwargs into the settings object.
     settings.peering.clusterwide = clusterwide
@@ -219,20 +223,19 @@ async def spawn_tasks(
     posting.settings_var.set(settings)
 
     # A few common background forever-running infrastructural tasks (irregular root tasks).
-    tasks.append(asyncio.create_task(
+    inner_tasks.append(asyncio.create_task(
         name="stop-flag checker",
         coro=_stop_flag_checker(
             signal_flag=signal_flag,
             stop_flag=stop_flag)))
-    tasks.append(asyncio.create_task(
+    inner_tasks.append(asyncio.create_task(
         name="ultimate termination",
         coro=_ultimate_termination(
             settings=settings,
             stop_flag=stop_flag)))
-    tasks.append(asyncio.create_task(
+    outer_tasks.append(asyncio.create_task(
         name="startup/cleanup activities",
         coro=_startup_cleanup_activities(
-            root_tasks=tasks,  # used as a "live" view, populated later.
             ready_flag=ready_flag,
             started_flag=started_flag,
             registry=registry,
@@ -242,7 +245,7 @@ async def spawn_tasks(
             memo=memo)))  # to purge & finalize the caches in the end.
 
     # Kill all the daemons gracefully when the operator exits (so that they are not "hung").
-    tasks.append(aiotasks.create_guarded_task(
+    outer_tasks.append(aiotasks.create_guarded_task(
         name="daemon killer", flag=started_flag, logger=logger,
         coro=daemons.daemon_killer(
             settings=settings,
@@ -250,7 +253,7 @@ async def spawn_tasks(
             operator_paused=operator_paused)))
 
     # Keeping the credentials fresh and valid via the authentication handlers on demand.
-    tasks.append(aiotasks.create_guarded_task(
+    authn_tasks.append(aiotasks.create_guarded_task(
         name="credentials retriever", flag=started_flag, logger=logger,
         coro=activities.authenticator(
             registry=registry,
@@ -261,7 +264,7 @@ async def spawn_tasks(
 
     # K8s-event posting. Events are queued in-memory and posted in the background.
     # NB: currently, it is a global task, but can be made per-resource or per-object.
-    tasks.append(aiotasks.create_guarded_task(
+    inner_tasks.append(aiotasks.create_guarded_task(
         name="poster of events", flag=started_flag, logger=logger,
         coro=posting.poster(
             settings=settings,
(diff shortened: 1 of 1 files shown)


nolar:
Closing in favour of

* #1286 

This PR **failed conceptually**: while the tasks graceful exiting works as expected (from inside out; for stop-flag waiter and SIGINT/SIGTERM signals), the asyncio cancellation does not: when we cancel the top-level operator, it immediately cancels all tasks at all levels, so there is no layering or gradual shutdown. As such, this is only a partial solution.

Moreover, this solution seems complicated. So far, in 6-7 years of Kopf's life, this only required once for authentication here. It is easier to solve that particular problem without over-generalisation. The newer solution, however, has some space for extensibility.

PR #1295, opened 2026-03-27
Isolate KopfRunner's handlers from the caller's handlers or neighboring runners

nolar:
**BLOCKED, see below.**
**ABANDONED in faviour of KopfCLI, the true subprocess runner with full isolation. See #1292.**

That was the intended behaviour: the `KopfRunner` loads the handlers from the files and modules specified on the command line, the same as `kopf run` does. But the behaviour was broken: in addition to the specified files and modules, the handlers from the caller were leaked into the operator. Moreover, the handlers of different runners were accumulated in the same default registry over time, affecting each other.

That wasn't a big problem likely because (1) the test runners are rarely used; (2) they are loading the same handlers of the same operator anyway, differing only in the behavior of the runner's block.

Example:

```python
import kopf.testing

@kopf.on.event('pods')
def unintentional_handler(**_): ...

with kopf.testing.KopfRunner(['run', '--verbose', 'file1.py', '-m', 'module1']):
    # unintentional_handler will be in effect in addition to file1/module1 handlers
    ...

with kopf.testing.KopfRunner(['run', '--verbose', 'file2.py', '-m', 'module2']):
    # unintentional_handler will be in effect in addition to file2/module2 handlers
    # handlers from file1.py and module1 will also be here in the 2nd runner
    ...
```

This fix isolates the runner from the caller of the runner and from other runners, so that the runner runs in its own isolated environment (registry).

If the **old behaviour** is needed (not recommended), pass the default registry of the caller explicitly:

```python
import kopf.testing

@kopf.on.event('pods')
def desired_handler(**_): ...

with kopf.testing.KopfRunner(['run', '--verbose', 'file.py', '-m', 'module'],
                             registry=kopf.get_default_registry()):
    ...
```

--- 

**BLOCKER:** Works with files. Does not work with modules as `-m mod`: if the module was imported in the caller, it is NOT reloaded in the runner by Python, so the handler registration does not happen again, so the isolated registry remains empty.

The only solution seems to be running as a fair subprocess or to tweak the module loading (e.g. force reloading of specified modules one every run, potentially causing double-execution of the top-level code, double class declaration and invalidating the previous classes, and other unpleasant side effects).

Sub-processes seem to be the closest to the original intention — a fully isolated CLI runner as in `kopf run`. For non-isolated runners, the new `KopfTask` & `KopfThread` can replace the old behaviour of `KopfRunner`.

But this breaks the backwards compatibility.

**ANOTHER solution:** keep it as is, non-isolated, with all the side effects, but backwards compatible. Declare it as deprecated (issue a `DeprecationWarning` on creation). Replace it with a new runner `KopfProcess` to behave properly with all the file/module isolation — in line with `KopfTask` & `KopfThread`.

### DIFF
--- a/docs/testing.rst
+++ b/docs/testing.rst
@@ -42,6 +42,34 @@ exit code and output are available to the test (for additional assertions).
     the same as if it were executed with ``kopf run``.
 
 
+Handler isolation
+-----------------
+
+:class:`kopf.testing.KopfRunner` is isolated from the caller's environment
+by default: it creates its own registry and settings, so only the handlers
+from the specified files and modules are loaded --- the same as ``kopf run``
+would behave on the command line.
+
+This means that any ``@kopf.on.*`` handlers defined in the test file
+or elsewhere in the caller's code are not visible to the runner
+and do not affect the operator being tested.
+
+If the caller's handlers must be passed to the operator (not recommended), pass
+the caller's default registry explicitly from :func:`kopf.get_default_registry`
+--- this is where the handlers go to by default:
+
+.. code-block:: python
+
+    import kopf
+    from kopf.testing import KopfRunner
+
+    with KopfRunner(
+        ['run', '--verbose', 'example.py'],
+        registry=kopf.get_default_registry(),
+    ) as runner:
+        ...
+
+
 Mock server
 ===========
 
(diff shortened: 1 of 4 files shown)


nolar:
ABANDONED in faviour of KopfCLI, the true subprocess runner with full isolation. See https://github.com/nolar/kopf/pull/1292.

PR #1302, opened 2026-04-15
Fix documentation for default event logging behaviour

### LINKED ISSUE #1301: Docs: Logging events documentation is incorrect
### Long story short

In https://docs.kopf.dev/en/stable/configuration/#logging-events

```
By default, all log messages made by the handlers on their logger are also posted as Kubernetes events. This can be disabled if it is not desired, e.g. to keep the events list clean, so that only the explicit event-posting calls are posted:
```

This appears to no longer be the case since 1.44.0

https://github.com/nolar/kopf/commit/3872f3c15be41d41d9faca3cef86322c50287530

### Kopf version

1.44.0

### Kubernetes version

N/A

### Python version

N/A

### Code

```python

```

### Logs

```none

```

### Additional information

_No response_

markafarrell:
Fixed documentation for default behaviour of logging events

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -151,10 +151,9 @@ These two settings also affect :func:`kopf.event` and related functions:
 :func:`kopf.info`, :func:`kopf.warn`, :func:`kopf.exception` ---
 even if they are called explicitly in the code.
 
-By default, all log messages made by the handlers on their ``logger`` are also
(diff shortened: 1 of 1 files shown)


nolar (approved):

nolar:
Thanks for noticing and fixing that!

